PR #7, opened 2015-01-19
Fixed accidentally global 'browserslist' var.

papandreou:
Caught this one because the project I wanted to use browserslist in uses `mocha --check-leaks`, so you might want to turn that on here as well :)


MoOx:
A linter should be added. Like jshint or eslint (I recommand this one)

ai:
I uses `jshint` https://github.com/ai/browserslist/blob/master/gulpfile.js#L6

ai:
Maybe I miss some options to check this errors?

ai:
Released in 0.1.2.

PR #15, opened 2015-02-26
Fix accidental global

Munter:
Leaking this causes mocha to throw errors


ai:
Thanks. Released in 0.3.1.

PR #21, opened 2015-04-11
DynJS compatibility

halfninja:
DynJS (http://dynjs.org/) fails on this code because this sort function sometimes returns floating point values, and it's a bit more strict than other engines about requiring integer values. This change makes the module work in DynJS, and other existing JS engines will also be fine with it.


ai:
Awesome.

Sorry, I right now in trip. I will check this PR after April 15.

ai:
Released in 0.3.3. Autoprefixer will use it in 5.2 release at the end of this month.

PR #25, opened 2015-06-17
Update actual example in README test

xavhan:


ai:
Thanks to keep 80 column limit :).

PR #40, opened 2016-01-01
Add the possibility to query for browser capabilities

DanReyLop:
**New queries:**
* `support <feature>` : Returns all the browser versions that support or partially support the given feature
* `supports <feature>` : Same as above
* `full support <feature>` : Returns all the browser versions with complete support for the given feature
* `fully supports <feature>` : same as above

**Rationale:**
To put it simply: If my webapp uses WebSockets, there is no point on generating CSS prefixes for IE8. I'ts a common use case to try to make the webapp compatible with as much browsers as possible, but if it uses a new technology, trying to run it in older browsers would break anyway. So, for the WebSockets example, one could query browserlist with `"supports websockets"` and that would match all the browser versions that support the feature.

I found it useful for me, of course I'm open to suggestions or changes, but I think this query adds more flexibility to the library.

The query uses the caniuse database. `"support <feature>"` matches partial and full support (the most common use case, since partial support of most features just means that a vendor prefix is needed or that some advanced usage of the feature is not possible), and `"full support <feature>"` matches only full support.

**Note:** This would be more useful if there was a way to compute queries with the boolean "AND" operation, such as `"supports websockets, and supports local-storage, and > 1%"`. Maybe that will be my next PR, if I manage to keep the logic reasonably simple.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,8 +11,8 @@ major browser, or have a usage of over 10% in global usage statistics:
 
 ```js
 browserslist('last 1 version, > 10%');
-//=> ['and_chr 46', 'chrome 46', 'chrome 45', 'edge 12', 'firefox 41'
-//    'ie 11', 'ie_mob 11', 'ios_saf 9', 'opera 32', 'safari 9']
+//=> ['and_chr 47', 'chrome 47', 'chrome 46', 'edge 13', 'firefox 43',
+//    'ie 11', 'ie_mob 11', 'ios_saf 9.0-9.2', 'opera 34', 'safari 9']
 ```
 
 Multiple criteria are combined as a boolean OR. In other words, a browser
--- a/README.md
+++ b/README.md
@@ -49,8 +49,12 @@ You can specify the versions by queries (case insensitive):
 * `Firefox <= 20`: versions of Firefox less than or equal to 20.
 * `Firefox ESR`: the latest [Firefox ESR] version.
 * `iOS 7`: the iOS browser version 7 directly.
+* `support css-gradients`: all browser versions that support to some extent the given feature (in this case, CSS Gradients)
+* `full support css-gradients`: all browser versions that _fully_ support the given feature (in this case, CSS Gradients)
 * `not ie <= 8`: exclude browsers selected before by this query.
   You can add `not ` to any query.
+
+To see the list of features that you can query for, visit http://caniuse.com
 
 Blackberry and Android WebView will not be used in `last n versions`.
 You should add them by name.
--- a/index.js
+++ b/index.js
@@ -389,12 +389,14 @@ browserslist.queries = {
             }
             for ( var name in data ) {
                 for ( var version in data[name] ) {
-                    // See caniuse-db/CONTRIBUTING.md for details on the support string format.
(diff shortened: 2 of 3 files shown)


ai:
Hm. Interesting feature. The only question is: maybe we should create other tool for it?

What is your use case for it?

ai:
Hm. Seems like we already have special library for it (sorry ^_^): https://www.npmjs.com/package/caniuse-api

DanReyLop:
My use case is that I have a webapp that uses WebSockets, and I want it to work in as much browsers as possible, so currently my only option to use AutoPrefixer would be something like:
`postcss([ autoprefixer({ browsers: ['> 0%] }) ])`
But that would create a lot of unneccesary prefixes, since the webapp is only usable in browsers with WebSocket support. So a cleaner way to do that would be:
`postcss([ autoprefixer({ browsers: ['supports websockets'] }) ])`
I don't think that's a strange use case, but maybe I'm wrong :)

You are right, caniuse-api actually uses browserlist to narrow down results, I'll try to send a PR to caniuse-api to add the possibility to actually list the browser versions that support a feature so it can be used with autoprefixer. It's written in ES6, won't be easy :(

ai:
I think you can use `caniuse-api` right now. Just add code to convert `caniuse-api`’s object to Autoprefixer list.

I know, that it is a extra code. But I think a llot about your suggestion and I think it will be bad for ecosystem. Let’s me explain:

We have Browserslist in Autoprefixer, especially because we think that `> 0%` is a vey bad way. Prefixes should be actual. There is no sense to support Chrome 4, for example. Default selection `> 1%, last 2 versions, Firefox ESR` is enough for most use cases.

This is why `supports websockets` is like a `> 0%`. It will select a lot of browsers without a user base. `supports websockets` will select **all** Chrome, including Chrome 5 with 0,01% user base.

ai:
Why `> 1%, last 2 versions, Firefox ESR` is not enough for you?

DanReyLop:
I think the defaults would be enough for my current use case, but I was trying to think beyond the current project at hand. From my point of view, if a webpage can be made to run in Chrome 4 without extra effort, it should be done. `> 0%` is _bad_ because it produces unnecessary CSS code, but I personally find the few extra bytes worth it if that means that a few users will see my webapp perfectly instead of a broken mess. 0,1% of the total population of the world is still a lot of people. So, I see `support <whatever>` as a small optimization over `> 0%`. Of course, it's clear that our opinions differ in that case, and that's not a bad thing.

PS: After seeing the code of caniuse-api, I think will create a small wrapper that uses both `browserslist` and `caniuse-api`, it will be easier and more practical than bloating an existing library (and maybe someone will find it useful on its own). I don't want to dive into ES6 just yet :D

ai:
If you will release this `caniuse-api` to `browserslist` wrapper, I can add it to Browsersllist docs. Please, send PR.

ai:
And sorry for rejecting PR, you made good work. It is only question of project philosophy.

DanReyLop:
Don't worry, I know that for every PR it's a chance that it will be rejected, I just feel that it's better to discuss about a feature (specially a really small one like this) over the actual working code than just proposing the change. I will release it, I'll ping you when it's ready :)

Issue #4, opened 2015-01-08
Add option to specify custom analytics data


MoOx:
caniuse.com UI offers this option. Maybe we should get it too. Don't you think ?

ai:
You can already use `> 5% in US`. You ask about it?

MoOx:
Hum. Good to know. Sorry to miss that. Renaming the issue.

ai:
So I close issue if Browserify already can do it? :)

MoOx:
You mean browserslist.
It cannot import custom analytics right now am I correct?

ai:
You mean usage statistics from some country? You can do it right now by:

```js
# Take browsers used by 5% of global users, 1% of Russian and Ukrainian users
browserslist('> 5%', '> 1% in RU', '> 1% UA')
```

MoOx:
Non I mean custom google analytics data like on caniuse.com

![screen shot 2015-01-10 at 14 54 42](https://cloud.githubusercontent.com/assets/157534/5691799/c7e692bc-98d8-11e4-8d73-2d397bbe83a4.png)

ai:
Ouh, I understand. Yeap, we should implement it.

@Fyrd is there some examples, for you parse GA data in canisue.com?

ai:
I suggest this API:

```js
browserslist('> 5% in my GA', { ga: gaFileContent });
```

Does '5% in my GA' sounds normal? I want some separated selector to be able to combine global statistics and your GA statistics.

ai:
Done in https://github.com/ai/browserslist/pull/41 by @DanReyLop

PR #41, opened 2016-01-02
Added the possibility to query against custom browser usage data

DanReyLop:
See issue #4.

**New query:**
`> X% in my stats`

**Usage:**
``` js
browserslist("> 5% in my stats", {ga: "path/to/usage_file.json"})
```
Also you can provide the usage info directly, like this:
``` js
var usage = {"ie": {"7":0, "8":4.5, ...}, "chrome": {"34":23.7, ...}, ...}
browserslist("> 5% in my stats", {stats: usage})
```
The expected file format is the same as `require('caniuse-db/region-usage-json/' + country).data`

The code used to parse the data directly from Google Analytics is over 1000 lines long and entangled with other parts of the caniuse.com JS code, so I just implemented a silly workaround:
* Go to http://caniuse.com and import your custom Google Analytics data, as you would normally.
* Paste this snippet into the navigation bar (tested in Chrome):
```js
javascript:(function(){var e=document.createElement('a');e.setAttribute('href', 'data:text/plain;charset=utf-8,'+encodeURIComponent(JSON.stringify(JSON.parse(localStorage['usage-data-by-id'])[localStorage['config-primary_usage']])));e.setAttribute('download','ga.json');document.body.appendChild(e);e.click();document.body.removeChild(e);})()
```

That will give you the JSON file needed by browserslist.

Tell me if the functionality is OK and I'll write some tests for it and update the documentation. Maybe we should adapt the CLI script to support this functionality, or the use of an environment variable to specify the usage file.


ai:
Great work. I will accept it when I come back to home

DanReyLop:
Please don't accept it as-is. I need to add some tests and also I didn't even check if the current tests pass (seems like they don't, for some reason). Now that I know that you are on board with this patch, I will finish it and document it as soon as I have time (tomorrow).

ai:
Yeap. We need tests and docs changes. Can I ask you to do it, because I am bad in English writing.

I can help you with tests, if you have any questions. You can take some example stats from some small site.

DanReyLop:
Done :)

ai:
Amazing work.

DanReyLop:
Well, that was fast :)
Thanks for the publicity on Twitter :D
Loving the npm ecosystem so far, the tools are so small that anyone can hack into them in no time.

PR #48, opened 2016-02-08
Add extra fs safeguard checks in readConfig

MeoMix:
Since readConfig calls existsSync and statSync it would be prudent to check for their existence first.

JSPM has a fledgling library implementing fs, https://github.com/jspm/nodelibs-fs/blob/master/fs.js, which causes a check for just `readFileSync` to be insufficient.


ai:
Released in 1.1.2.

MeoMix:
Thanks for the quick acceptance and release.

PR #53, opened 2016-04-12
Add query for include or exclude all version of the browsers.

hikouki:
I'd like to add query for include or exclude all version of the browsers.

### DIFF
--- a/index.js
+++ b/index.js
@@ -434,8 +434,17 @@ browserslist.queries = {
             }
             return [data.name + ' ' + version];
         }
-    }
+    },
 
+    allVersions: {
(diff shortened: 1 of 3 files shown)


ai:
Sorry. I know that many developers want it, but it will promote very bad behavior. I can't find real use case for all browsers. Of course l, you site doesn't support IE 3 or Chrome 3. Developers should think better about browsers support policy and avoid "all versions".

ai:
If you really need it, you can use unofficial "ie >= 0". The long and wierd syntax of it protect developers from using it too often.

hikouki:
I see. I  understood it.
Thank you for reply. :smile:

PR #57, opened 2016-05-19
Fixed wrong numbers in example queries

mdix:


ai:
You are late 😆 Already fixed in PR just few seconds ago

mdix:
Darn.. ok, perfect then.

ai:
But, awesome that you send PR so quick, thanks. Anyway, more contributers is better

PR #60, opened 2016-06-03
Fix link to caniuse.com in README

dnahodil:
Link works (previously was relative to this page)
Link display (capitalisation) is consistent with other links in this document


ai:
Thanks!

PR #79, opened 2016-11-21
Use plural "last 2 versions" to match other examples

msheakoski:


msheakoski:
The change isn't displaying for me in the "Files changed" tab. If it's not working correctly, what I did was edit line 207 of README.md and changed it to: `browserslist "> 1%, last 2 versions"` (added an "s" to "version").

Issue #74, opened 2016-10-13
Add package.json support


ai:
``` js
{
  …
  "browserslist": [
    "last 1 version",
    "ie 10"
  ]
  …
}
```

ai:
Done by @stepankuzmin in #85

PR #85, opened 2016-12-14
Add package.json support #74

stepankuzmin:
Add `package.json` support according to [this task](http://cultofmartians.com/tasks/browserslist-package.html).


ai:
Great. Thanks. Will accept it after a dinner.

If you want some feedback: it is dangerous to check just `.toThrow()`, because maybe you have a issue in `new Error` code :D. It is a good practice to always check error message by `toThrowError(/contains both/)`

stepankuzmin:
Right! Thanks for the review!

haisarov:
ggg

PR #86, opened 2016-12-14
Add load statistic from browserslist-stats.json

AleshaOleg:


README.md
```diff
@@ -157,6 +157,7 @@ If you have a website, you can query against the usage statistics of your site:
     ```js
    browserslist('> 5% in my stats', { stats: 'path/to/the/stats.json' });
     ```
+5. Import statistic from `browserslist-stats.json` in project folder
```
ai: It is not a separated step.  It is just part of 4 step. Also feel free to rewrite section to make it more clear.

index.js
```diff
@@ -72,9 +72,34 @@ var browserslist = function (selections, opts) {
...
+
     if ( opts.stats || process.env.BROWSERSLIST_STATS ) {
+        statsExist = true;
+    } else {
+        (function searchStatsFile(startPath) {
+            if (!fs.existsSync(startPath)) {
+                return;
+            }
+
+            var files = fs.readdirSync(startPath), filename, stat, currFile;
```
ai: Mixed variable difinition with values `var a = 1` and without `var a` is prohibited in this project :).

index.js
```diff
@@ -72,9 +72,34 @@ var browserslist = function (selections, opts) {
         selections = selections.split(/,\s*/);
     }
 
+    var statsFile, statsExist;
+
     if ( opts.stats || process.env.BROWSERSLIST_STATS ) {
+        statsExist = true;
+    } else {
+        (function searchStatsFile(startPath) {
```
ai: Too tricky code. It is better to use `while` from config loader without a recursion. Recursion is always hard to understand.

index.js
```diff
@@ -72,9 +72,34 @@ var browserslist = function (selections, opts) {
...
+            if (!fs.existsSync(startPath)) {
+                return;
+            }
+
+            var files = fs.readdirSync(startPath), filename, stat, currFile;
+            for (var i = 0; i < files.length; i++ ) {
```
ai: This project code style need to have space after `(` in `if`/`for`/`while`.

test/fixtures/browserslist-stats.json
```diff
@@ -0,0 +1,7 @@
+{
+    "ie": {
+        "6": 0.01,
+        "7": 0.4,
+        "8": 1.5
+    }
+}
```
ai: No new line at the end of file.

test/custom.test.js
```diff
@@ -40,3 +41,9 @@ it('works alongside global usage query', () => {
     var list = browserslist('> 10% in my stats, > 1%', { stats: usage });
     expect(list.length > 1).toBeTruthy();
 });
+
+it('take stats from browserslist-stats.json', () => {
+    var data = JSON.parse(fs.readFileSync(statsFile));
```
ai: You pass a stats by `stats` argument in this test case. It doesn’t test that you look up in directories.

Look how I test a config file.

index.js
AleshaOleg: @ai but as I understood from task, it should to search file in any folder of project folder. Or same `while` loop as in config search will be enough?

ai: @AleshaOleg if we process `/a/b/c/file.css` we should check stat file in `/a/b/c`, `/a/b`, `/a` and `/`.

And we already do this “recursion” search in config loading (but without recursion in code).

README.md
```diff
@@ -151,8 +151,9 @@ If you have a website, you can query against the usage statistics of your site:
    var e=document.createElement('a');e.setAttribute('href', 'data:text/plain;charset=utf-8,'+encodeURIComponent(JSON.stringify(JSON.parse(localStorage['usage-data-by-id'])[localStorage['config-primary_usage']])));e.setAttribute('download','stats.json');document.body.appendChild(e);e.click();document.body.removeChild(e);
     ```
 3. Save the data to a file in your project.
-4. Give it to Browserslist by `stats` option
-   or `BROWSERSLIST_STATS` environment variable:
+4. Give it to Browserslist by `stats` option,
+   `BROWSERSLIST_STATS` environment variable
+   or import from `browserslist-stats.json` in project folder:
```
ai: I think “save to `browserslist-stats.json`” will be much more clear, than “import” (also feel free to put this way as first way).

test/custom.test.js
```diff
@@ -40,3 +40,7 @@ it('works alongside global usage query', () => {
     var list = browserslist('> 10% in my stats, > 1%', { stats: usage });
     expect(list.length > 1).toBeTruthy();
 });
+
+it('take stats from browserslist-stats.json', () => {
+    expect(browserslist('> 5% in my stats')).toEqual(['ie 8']);
```
ai: Why this test are working? You didn’t pass a `file` option, so Browserslist should not know where to find.

index.js
```diff
@@ -519,4 +541,6 @@ browserslist.queries = {
     }
 }());
 
+browserslist();
```
ai: Why you call empty function here?

index.js
```diff
@@ -72,9 +72,11 @@ var browserslist = function (selections, opts) {
         selections = selections.split(/,\s*/);
     }
 
-    if ( opts.stats || process.env.BROWSERSLIST_STATS ) {
+    var statsFile;
+
+    var getStats = function () {
```
ai: Code become better but still too tricky. What do you think about this:

```js
var stats = opts.stats || process.env.BROWSERSLIST_STATS
if ( !stats ) {
  // search in files
}
if ( !stats ) {
  // show error
} else {
  // use custom stats
}
```

In this case you don’t need `getStats` function and could read code from top to bottom, without jumping in file.

README.md
AleshaOleg: @ai just updated, please review README again.

index.js
AleshaOleg: @ai ESLint thinks it's a complexity in this case, trying come up with a solution

ai: @AleshaOleg you can always change `complexity` settings, you you understand that they are wrong in this case.

Or move it to `var stats = browserslist.loadStats()`

Issue #111, opened 2017-01-25
CLI does not allow passing stats path nor environment


jrencz:
it only has an argument for `config`, but there's none for `stats` and no `env`

`browserslist` complains if stats are required and the only way to provide it is `BROWSERSLIST_STATS` env.

ai:
Released in 1.7

PR #112, opened 2017-01-25
Allow passing stats and env when `—-config` is given (#111)

jrencz:


ai:
Thanks. I forgot about CLI.

ai:
I will release it tomorrow or aftertomorrow

jrencz:
How do you feel about using more declarative tool for parsing arguments, let's say yargs? I may reorganise CLI with that

ai:
@jrencz CLI tool for Browserslist are not really popular. So making extra dependency for rare feature is not good way :D.

PR #117, opened 2017-02-12
Reimplement `eachParent`'s traversal

akx:
As discussed in #116.

Previously `eachParent` sliced the path all the way up to the empty string. Practically this meant that calls to `eachParent` ended up _also_ taking into consideration a `package.json` in the current working directory.

This implementation is based on [the traversal code in Babel](https://github.com/babel/babel/blob/master/packages/babel-core/src/transformation/file/options/build-config-chain.js#L60).


ai:
Thanks. Released in 1.7.2.

PR #120, opened 2017-02-15
Restore eachParent traversal behavior

akx:
The eachParent traversal modification in bc86e87e0e958e caused eachParent to skip the first iteration, as it were, compared to earlier behavior.

I.e. if you passed in `/foo/bar/quux`, 1.7.1 and before iterated over
 
* `/foo/bar/quux`
* `/foo/bar`
* `/foo`
* (erroneously) ``

1.7.2 and later (with the modification in bc86e87e0e958e) iterated over 

* `/foo/bar`
* `/foo`
* (correctly) `/`

given the same parameter.

This restores the 1.7.1 behavior, but with the `` bug corrected.

### DIFF
--- a/index.js
+++ b/index.js
@@ -56,10 +56,10 @@ function eachParent(file, callback) {
     if ( file === false ) return undefined;
     if ( typeof file === 'undefined' ) file = '.';
     var loc = path.resolve(file);
-    while (loc !== (loc = path.dirname(loc))) {
+    do {
         var result = callback(loc);
         if (typeof result !== 'undefined') return result;
-    }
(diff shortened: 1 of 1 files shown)


akx:
Whoa, that was a fast merge 😅

ai:
I made a release commit and will publish it after Travis CI will finish.

ai:
Released in 1.7.3

akx:
Great! Sorry about introducing that bug! 😔

ai:
I think nobody got it, because all main Browserslist users send a file to `path`

PR #123, opened 2017-02-17
Don't use locale-aware comparison for browser names

akx:
I used [`0x`](https://github.com/davidmarkclements/0x) to look at the flamegraphs for (you guessed it!) a CSS build, and found out a lot of time was spent sorting the browser list.

It might be a good idea to add an option to disable the sorting, since many use cases don't really need a prettily sorted list anyway?

Anyway, I don't think there's a good reason to use the collation-aware `String.localeCompare()` method for comparing browser names, which are always in ASCII.

## Performance impact

Using this highly synthetic benchmark which simply calls the `browserlist()` function 10,000 times, this branch is **28% faster** than 1.7.3.

```shell
$ time node -e 'bl=require(".");for(i=0;i<10000;i++){bl([">1%"])}'
```
* `master` (19b3610): 1.44 seconds
* `no-locale-compare`: (196c739): 1.12 seconds

### DIFF
--- a/index.js
+++ b/index.js
@@ -124,6 +124,12 @@ function generateFilter(sign, version) {
     }
 }
 
+function compareStrings(a, b) {
+    if (a < b) return -1;
+    if (a > b) return +1;
+    return 0;
+}
+
 /**
  * Return array of browsers by selection queries.
  *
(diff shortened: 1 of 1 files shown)


ai:
Wow, great! We could release it in patch version.

I am doing to drink right now. So I will accept PR and release it only tomorrow. But it looks awesome, thanks!

akx:
We're already on it here! :) 

За здоровье! 😄 

photo on 17-02-2017 at 16 09

ai:
Thanks. Released in 1.7.4.

PR #124, opened 2017-02-22
fix --env

tuures:
should set `opts.env` instead of `opts.config`


ai:
Thanks! I missed it :).

ai:
Released in 1.7.5.

PR #132, opened 2017-03-24
[WIP] Initial cli tests

vvscode:


cli.js
```diff
@@ -4,6 +4,7 @@ var browserslist = require('./');
 var pkg          = require('./package.json');
 var args         = process.argv.slice(2);
 
+
```
ai: Oops, you add unnecessary newline ;)

npm-debug.log
```diff
@@ -0,0 +1,46 @@
+0 info it worked if it ends with ok
```
ai: You commented also unnecessary `npm-debug.log`. I added it to `.gitignore` a4342a3, but you should check pull request diff next time.

test/cli.test.js
```diff
@@ -0,0 +1,73 @@
+const pkg = require('../package.json');
+const CLI_PATH = '../cli';
+
+let write;
```
ai: `let` and `const` doesn‘t work in Node.js 4

ai:
I added first review, fix it first. Tomorrow I will look into tests (I am drunk right now).

But right now I think mocking is too complicated. There is a better way:

1. Call real CLI by `require('child_process').spawn`.
2. Compare it with `browserslist()` function result. In this case you don’t need to mock CLI data.

vvscode:
But such tests will be slow - is it ok? also they will work with network ( if I understand how browserlist works ). is it normal for tests?

ai:
> But such tests will be slow - is it ok? 

It is OK :).

> also they will work with network

Nope, Browserslist doesn’t use network. It [has](https://github.com/ai/browserslist/blob/master/package.json#L13) own copy of Can I Use database.

PR #133, opened 2017-03-25
Tests for cli.js

zhulduz:


ai:
Great and fast work :+1:.

If you want to learn something new from this task:

1. `spawn.stdout.on('data')` is asynchronous callback. It could be executed after a second. So you should return Promise from test.
2. It is better to wait `spawn.stdout.on('close')` because big output could call several `data` events.

Here is a example, how I solved this notices: 1a961a4

ai:
What is your Twitter and VK accounts for announce?

zhulduz:
@ai thanks for the feedback) Now it seems much better.

https://twitter.com/zhankenova, I don't have VK account(

PR #152, opened 2017-05-27
Support "all" queries

### LINKED ISSUE #151: Idea: targeting all versions of a browser / all browsers
Sometimes I use a browser list for individual parts of my app, like I do in [PostCSS Normalize](https://www.npmjs.com/package/postcss-normalize).

For instance, I might have a fix for _all versions_ of IE. Right now, to mark this up I write `ie > 0`, but could it instead be `ie`?

In another instance, I might have a fix for _all browsers_. Right now, to mark this up I write `> 0%`, but could it instead be `all`?

jonathantneal:
Adds support for queries like `all ie`, `all ff`, and even `all`

Resolves #151

### DIFF
--- a/index.js
+++ b/index.js
@@ -619,6 +619,27 @@ browserslist.queries = {
         }
     },
 
+    all: {
+        regexp: /^all(?:\s+(\w+)\s*)?$/i,
+        select: function (context, name) {
+            if (name) {
+                var optdata = browserslist.checkName(name);
+
+                return optdata.released.map(function (v) {
+                    return optdata.name + ' ' + v;
+                });
+            } else {
+                return Object.keys(caniuse).reduce(function (all, eachname) {
+                    var eachdata = browserslist.byName(eachname);
+                    if ( !eachdata ) return all;
+                    return all.concat(eachdata.released.map(function (v) {
+                        return eachdata.name + ' ' + v;
+                    }));
+                }, []);
+            }
+        }
+    },
+
     esr: {
         regexp: /^(firefox|ff|fx)\s+esr$/i,
         select: function () {
--- a/test/browser.test.js
+++ b/test/browser.test.js
@@ -4,6 +4,11 @@ var originData = browserslist.data;
 
 beforeEach(() => {
     browserslist.data = {
+        edge: {
+            name:     'edge',
+            released: ['12', '13', '14', '15'],
+            versions: ['12', '13', '14', '15']
+        },
         ie: {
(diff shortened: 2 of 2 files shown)


test/browser.test.js
```diff
@@ -16,6 +21,15 @@ afterEach(() => {
     browserslist.data = originData;
 });
 
+it('selects all versions of browser', () => {
+    expect(browserslist('all ie')).toEqual(['ie 11', 'ie 10', 'ie 9']);
+});
+
+it('selects all versions of all browsers', () => {
```
11bit: I think breaking line after 'edge 13' doesn't correspond to other code style. What about this version:
```
it('selects all versions of all browsers', () => {
    expect(browserslist('all')).toEqual([
        'edge 15', 'edge 14', 'edge 13', 'edge 12', 'ie 11', 'ie 10', 'ie 9'
    ]);
});
```

jonathantneal:
Thank you for the review. I have made all of the changes you requested.

index.js
```diff
@@ -619,6 +619,28 @@ browserslist.queries = {
         }
     },
 
+    all: {
+        regexp: /^all(?:\s+(\w+)\s*)?$/,
```
jonathantneal: Please see https://github.com/ai/browserslist/pull/152/commits/24f5c1bbec7dde7c7032361a9ddac9dbe59c1732#diff-168726dbe96b3ce427e7fedce31bb0bcR623

index.js
```diff
@@ -619,6 +619,28 @@ browserslist.queries = {
...
+        select: function (context, name) {
+            if (name) {
+                var optdata = browserslist.checkName(name);
+
+                return optdata.released
+                .map(function (v) {
```
jonathantneal: Please see https://github.com/ai/browserslist/pull/152/commits/24f5c1bbec7dde7c7032361a9ddac9dbe59c1732#diff-168726dbe96b3ce427e7fedce31bb0bcR628

test/browser.test.js
jonathantneal: Please see https://github.com/ai/browserslist/pull/152/commits/24f5c1bbec7dde7c7032361a9ddac9dbe59c1732#diff-e5931c6e76a153b748f247af851fee40R30

11bit:
Thanks @jonathantneal, looks good now. The only question is what @ai will decide regarding this feature. As he commented in corresponding issue, generally it is a bad practice to select all browsers.

PR #170, opened 2017-08-25
New Feature: Major versions queries

oigewan:
**Feature:**
Fixes: https://github.com/ai/browserslist/issues/169

Add two new query types which allow users to specify which major versions of a browser/all browsers to allow.  Examples:

* `last 2 major versions`: all minor/patch releases of the current and previous major versions
* `last 2 iOS major versions`: all minor/patch releases of the current and previous major versions of iOS Safari

**Changes:**
- Added a `getLastMajorVersions` utility function for filtering the released versions of a browser to only include the last two major versions.
- Added and used a `curryMapName` utility method to keep the file size down.  There were plenty of select functions that ended by mapping over the resulting version numbers to prefix them with the browser name.  `curryMapName` accepts the browser name and curries a function which can be passed to `Array.prototype.map` in order to do the prefixing.
- Tests
- README update

cc: @ai

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,7 +78,9 @@ We recommend to write queries in `package.json`.
 You can specify the versions by queries (case insensitive):
 
 * `last 2 versions`: the last 2 versions for each browser.
+* `last 2 major versions`: all minor/patch releases of the current and previous major versions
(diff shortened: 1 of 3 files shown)


ai:
Looks good. I right now preparing for talk. Will check it in Sunday.

Issue #122, opened 2017-02-17
Enable sharing of browserlist among projects


morewry:
This project addresses the need to share a browerlist among tools, but I think it's equally important to be able to share a list among multiple modules. Tools such as stylelint and eslint enable sharing configurations with `extends` properties that correlate to the name of a preset installed as a Node.js. 

The biggest reason this is useful is that you would typically want to base your supported browsers on the traffic to _your_ website (rather than aggregate traffic to other websites). While you can generate a browserlist from your website stats and use it, it currently looks like you would have to make and update multiple copies of the list. Unless this is something enabled via babel-env-preset or another option. (It's not apparent to me, but I might be missing it).

Adding a feature that enables a user to have and use a browser list module would make this configuration even more useful than it is already.

e.g. in a `package.json`

```json
...
"browserlist": {
  "extends": "my-browserlist"
}
...
```

```shell
npm install --save-dev browserlist-preset-my-browserlist
```

morewry:
It looks like,

> You can specify direct path in BROWSERSLIST_CONFIG environment variables.

This could serve a similar purpose. It's preferable to avoid having to maintain specific paths, but it addresses the general need.

morewry:
> Save the data to a browserslist-stats.json file in your project.

If this has to be saved in every project, I'm back to needing a way to share it. :-)

ai:
Hm. Yeap, I like Stylelint/ESLint `extends` too. But I think in Browserslist case `BROWSERSLIST_CONFIG` will be enough.

There are few difference with Stylelint/ESLint:

* Browserslist config is smaller. In most cases it is just few lines.
* In linters you must to define your rules. In Browserslist it is better to have no `browserslist` config — defaults is better for most of project. I afraid that mobile web developers will start to use `last 2 iOS versions` and we will be back in `-webkit-` hell :(.
* I hope `browserslist` config will be used by non-JS tools. But `extends` force to use `npm`.

But there is no really important reason in that reasons. So let’s keep it open to get feedback, maybe people really want `extends` and I am wrong :D.

ai:
We have no feedback, so I will close issue

destroyerofbuilds:
I just came across this issue, and wanted to share my own setup in hopes it provides context on why supporting an `extends` like feature may be valuable.

At the company I work for our customers use a desktop application that ships with an embedded version of Chrome. Customers may be using any version of our desktop application released within the past few years, so they may be using one of several versions of Chrome.

[...]

PR #171, opened 2017-08-29
Support resolving shared config from external package

YellowKirby:
Adds support for a special "use" query which requires the corresponding package name and merges that config into the browserslist array.

I didn't add support for automatic `browserslist-config-` prefixing because that gets a little scary with scoped packages. Would be happy to add that in if you think it's necessary.

Related: #122


ai:
Good idea. Few questions:

1. Should we use `extends` instead of `use` keyword?
2. I think we should fall on wrong path instead of warn and ignore. Users miss warnings.
3. You should put new query code into others queries array.

YellowKirby:
> Should we use extends instead of use keyword?

Yeah, I think that is clearer and more consistent with how eslint, stylelint, etc. are configured.

> I think we should fall on wrong path instead of warn and ignore. Users miss warnings.

Makes sense. Updated.

> You should put new query code into others queries array.

Could you please clarify? I'm not entirely sure what new query code is. I added/updated some tests that use ranges/relative versions (`last 2 versions`) instead of everything referencing an explicit single version (`ie 11`). Is this what you're referring to?

Also any advice on getting the appveyor status to pass? It's timing out on Node 8 but I don't really understand what's failing.

Thanks!

ai:
Great!

Don't worry about AppVeyor, it is on me.

You implement new query by:

```js
queries = flatMap(queries, resolveExternalPackage)
```

But Browserslist implements all other queries by special array of regexps and callbacks:
https://github.com/ai/browserslist/blob/master/index.js#L484

If you will implement `extends` in this `QUERIES` array you will be able to remove some helper functions.

YellowKirby:
Got it, thanks @ai. Should be better now (I needed to move the thing that iterates over `QUERIES` into its own function so that I could call it recursively on the imported arrays). Let me know if there's anything else you'd like me to do.

ai:
There is last concern — security. What is somebody will accept Browserslist query from user and run it from user?

I agree that it is the rare case, but security is important.

What do you think?

My idea (it is only draft, I am open for any suggestions):

1. Extend path must start from `browserslist-config-*`.
2. Extend path should not contain `/../`.
3. Extend path should not contain `/node_modules/`.

This rules will reduce flexibility, but should increase security. Also, we can add `dangerousExtend` option to disable this rules.

YellowKirby:
Would it be easier to just have an optional to disable `extends` queries altogether? If a Browserslist config is being generated from user input then I would imagine the server wouldn't want to trust _any_ dynamic `require` statement.  Given how easy it is for anybody to publish an npm package under any name, I think having a whitelist for certain paths or package names would be fairly simple to circumvent.

Alternatively, could we use something like https://github.com/patriksimek/vm2 (even as just an `optionalDependency`) to sandbox anything that gets `require`d?

ai:
@YellowKirby I thought about disable option too. But the main problem is that users don’t read docs. So without explicit error nobody will think about security.

If we will block query, this error message will force users to read docs and found `dangerousExtend` option.

If we will have `disableExtends` options, nobody will find it, since nobody cares about security before hacker attack :(.

ai:
> Given how easy it is for anybody to publish an npm package under any name, I think having a whitelist for certain paths or package names would be fairly simple to circumvent.

But you need to install npm package first. So users can’t pass `extend browserslist-config-my-hack-npm-package` to server since server’s `package.json` has no `browserslist-config-my-hack-npm-package` in dependencies.

ai:
@YellowKirby if you want I can finish security check. Anyway, you did most of the work, it was really great.

YellowKirby:
@ai sorry I've been busy with some other projects and haven't had much time for this PR. If you want to take the security check stuff that'd be fine. Thanks!

ai:
:+1: I will finish and release it on weekend. Do you have Twitter to mentioned you in announce?

YellowKirby:
Awesome, thank you! I don't have a twitter account or much of an online presence =/

YellowKirby:
@ai: @brianchung808 added in the security stuff talked about in https://github.com/ai/browserslist/pull/171#issuecomment-327378141. Please let us know if there's anything else you'd like us to add/update.

Thanks!

ai:
Awesome! I will try to release it tomorrow on in the weekend. You did a really great work!

ai:
Released in 2.5

PR #190, opened 2017-11-01
Added a link to `browserslist-useragent`

pastelsky:


ai:
Sure, we really need this link here. But I think we need separated `Tools` sections. What do you think about tit?

ai:
Maybe we can put Tools section even before `Queries` since we don’™ have too many tools right now.

pastelsky:
Sounds good. Updated the PR. Let me know if I should change anything.

PR #194, opened 2017-11-08
Add support for "last electron versions" and "last electron major versions"

pop123123123:
I didn't add tests for latest versions of Electron because it depends on [electron-to-chromium](https://github.com/Kilian/electron-to-chromium). If tests are needed, consider making the e2c variable a property of the browserslist object, and then overwrite it.
I could have replace the entire **unreleased** function by a 
```js
return []
```
but it seemed cleaner this way.

### DIFF
--- a/index.js
+++ b/index.js
@@ -517,11 +517,28 @@ var QUERIES = [
     }
   },
   {
+    regexp: /^last\s+(\d+)\s+electron\s+major versions?$/i,
+    select: function (context, versions) {
+      var validVersions = getMajorVersions(Object.keys(e2c).reverse(), versions)
+      return validVersions.map(function (i) {
+        return 'chrome ' + e2c[i]
+      })
+    }
+  },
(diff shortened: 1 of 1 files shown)


ai:
1. Do we have `unreleased` versions in `e2c`?
2. We need tests since we have coverage check. You can just test `browsersist('last 2 Electron versions').length === 2`

ai:
Released in 2.9

Kilian:
Hi! :wave: Beta (= unreleased) releases for Electron are a very recent thing and are not yet described as such in the official versions lists. I'll see about adding that, and will then be able to add that status to `electron-to-chromium`!

ai:
@Kilian I am not sure that we really need them :D

Kilian:
@ai I'll keep that in mind, but they might be easy to add and then I'm sure someone will find a use!

PR #209, opened 2018-01-03
Support 'last 2 years' query

### LINKED ISSUE #208: Relative "since" query
https://github.com/ai/browserslist/pull/189 added the `since <year>` query which supports absolute years (i.e `2013`) 🎉 

Could the `since` query be expanded to support `since last <number> years` to target browsers without a hard-coded value like the `last <number> versions` query?

Edit: `last <number> years` may be a better query

wopian:
Expanding upon https://github.com/ai/browserslist/pull/189, this PR adds support for relative year queries with `last <number> years` to target all versions released within `<number>` amount of years. E.g [`last 3 years, not < 1%`](http://browserl.ist/?q=since+2016%2C+not+%3C+1%25)

Closes #208

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,6 +92,7 @@ You can specify the versions by queries (case insensitive):
 
 * `last 2 versions`: the last 2 versions for each browser.
 * `last 2 Chrome versions`: the last 2 versions of Chrome browser.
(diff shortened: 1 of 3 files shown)


ai:
Thanks. Released in 2.11.0.

PR #215, opened 2018-01-20
Change defaults: >0.3% instead of (>1% + last 2 versions)

LEW21:
This drops support for unused browsers - like QQ, Baidu, Blackberry, Opera Mobile, and IE 10; and adds back support for Chrome 49, which is the last release for Windows < 7 (still has 0,9% globally).

I'm still keeping last 2 Chrome and last 2 Firefox versions - as they're released really frequently, and it's possible that somebody haven't updated yet.

### DIFF
--- a/index.js
+++ b/index.js
@@ -256,8 +256,9 @@ browserslist.usage = {
 
 // Default browsers query
 browserslist.defaults = [
-  '> 1%',
-  'last 2 versions',
+  '> 0.3%',
+  'last 2 Chrome versions',
+  'last 2 Firefox versions',
   'Firefox ESR'
 ]
 


ai:
QQ and UC is extremely popular in China. We can't drop them.

Or give me better reasons why modern websites should not work for Chinese users 😉

ai:
Also it is very big major change. This type of changes (default which will affect > 10 M developers) must have better research and reasons.

LEW21:
This does not remove UC - it has 8.93% globally.

On the other hand QQ, according to caniuse, has 0.0% globally. But their source says it had 0.27% in Dec 2017 (and it's on its way to 0.3%), which suggests that they have out-of-date data.

Chrome 49 has 0.9% globally, and is not updateable for its users (because Chrome 50 won't work on their OSes) - I think that's a good reason to include it.

LEW21:
BTW since 2014, QQ is based on Chromium, so omitting it shouldn't break anything for apps that still support Chrome. Its latest version uses Chromium 53.

LEW21:
Additionally, according to browserl.ist, defaults have 87.56% global coverage, while my proposal has 91.87% global coverage.

ai:
Your proposal will block QQ browser or any new browser without big use base.

If you want to increase coverage, let's just change `>=1%`. Removing `last n versions` is against main goal of Autoprefixer and Browserslist (browsers diversity).

LEW21:
If these new browsers are based on existing browsers (like QQ or Baidu), they'll work correctly - and nowadays nobody starts to write a new browser from scratch. On the other hand, I don't think the goal should be to support explicitly all existing browsers - especially the ones that are not based on a modern engine, and have really small user base, and/or are dead like IE10, IE Mobile, Blackberry Browser.

Actually, the current query means that support for IE10 and IE11 will never be dropped - as there won't be a new IE version, IE10 and 11 will remain last 2 versions forever.

ai:
> they'll work correctly

Nope, QQ uses old Chromium, so even a prefixes are different from latest Chrome.

> I don't think the goal should be to support explicitly all existing browsers 

It is subjective. But I believe, that diversity is a key for progress.

> dead like IE10, IE Mobile, Blackberry Browser.

Removing dead browsers (by `not IE 10`) is negotiable. But I need explicit criterias for `dead` (doesn't supported and has less than 0.5% in global statistics?)

LEW21:
> Nope, QQ uses old Chromium, so even a prefixes are different from latest Chrome.

Latest QQ uses Chromium 53, while >0.3% includes Chromium 49, so it's going to be OK. And it's growing in popularity, so probably it'll get to the 0.3% level soon (or at least before Chrome 49 dies).

> Removing dead browsers (by not IE 10) is negotiable. But I need explicit criterias for dead (doesn't supported and has less than 0.5% in global statistics?)

This sounds OK, but it doesn't include browsers that are still supported, but their market share has fell to, or always was at, almost zero - like Opera Mobile fell, Opera Desktop probably will fall soon, and like all the open source browsers nobody heard about (Konqueror, Arora, rekonq, Epiphany, ...).

ai:
> Latest QQ uses Chromium 53, while >0.3% includes Chromium 49

I prefer to include QQ explicitly.

> but it doesn't include browsers that are still supported, but their market share has fell to, or always was at, almost zero - like Opera Mobile fell, Opera Desktop probably will fall soon, and like all the open source browsers nobody heard about 

One of the reason why I created Autoprefixer was that developers ignored these browsers 😉.

Opera was very popular is Russia and Norway. But nobody uses it in other countries. As result it has very low market share globally. But removing Opera support was like removing access for all Russians and Norwegians. It is really bad feeling when you see this behavior.

Also how we will get new browsers, if we will support only popular browsers? In this case we will have chicken and egg problem — nobody uses new browsers, because websites don't support this browser, nobody support this browser because users don't use it.

So, supporting this browsers is political decision. You should give me really good reasons to change this decision (I am Russian and Linux user and I know how bad is when websites ignore your browser).

I recommend you to focus on removing dead browsers and present good criteria and plan for it.

LEW21:
> I am Russian and Linux user and I know how bad is when websites ignore your browser

I am Polish and Linux user, I have some patches in both Arora and Konqueror, and I was trying to use Konqueror as my main browser in the past. Still, I believe that all browsers that do not have resources to keep up with modern web standards - which means implementing them themselves, or rebasing frequently on WebKit, Gecko or Chromium - should just die, as they're slowing down the development of the Internet.

> Also how we will get new browsers, if we will support only popular browsers? In this case we will have chicken and egg problem — nobody uses new browsers, because websites don't support this browser, nobody support this browser because users don't use it.

If people use feature detection - and polyfills are doing just that - then it's not necessary to support all the browsers explicitly. Also, nobody prevents new browsers from impersonating existing ones - like everybody does with that "Mozilla/whatever (Webkit like KHTML) etc" useragent strings, and implementing support for properties with other engine's prefix (like [Opera did](http://www.opera.com/docs/specs/presto2.12/css/aliases/) - and it's good, because they were meant to mean "syntax proposed by WebKit devs" and not "works only on WebKit"; OpenGL extensions also use this way of prefixing, NVidia implements many AMD-prefixed extensions, and vice versa).

I think the biggest use case for this library is to automatically determine the basic list of required features that are supported by the browsers, and then use polyfills only for the ones that are not yet supported. Then, if the new browser does not support features that are supported by 90% of market share, it's not really a browser worth helping.

awkaiser:
I believe current approach of `last 2 versions` default is correct.

With how easy it is to configure your own rules (by hand or via the numerous **browserslist-config-*** shared configs available on NPM today), why _wouldn't_ the default err towards inclusivity?

awkaiser:
However, including Chrome 49 is a fine point. It's below 1% but I suspect it will remain in use for quite some time.

ai:
@LEW21 I started to work under `not dead` query https://github.com/ai/browserslist/tree/dead

ai:
New defaults were released in 3.0

LEW21:
Nice :D

PR #227, opened 2018-02-09
feat: support multiple env in one section

evilebottnawi:
Close https://github.com/ai/browserslist/issues/226


ai:
Could you also add to test that `staging` env way parsed correctly (anyway I will be able to merge PR only hour later)

evilebottnawi:
@ai Also update README

test/fixtures/env-config/browserslist
```diff
@@ -6,6 +6,9 @@ ie 10
 ff 50
 chrome 55
 
-[production]
+[staging]
+edge 16
```
ai: This line means merging?

evilebottnawi: @ai yes, this is rare use case, but it is sometimes need

ai: From Ruby on Rails I had big experience in YAML. And all that flexibility to reduce few lines in config made bad things with the format.

So after YAML story, I think Browserslist config should be very simple. Let’s throw a error in this case. One header will reduce readability (it is easy to miss second section below).

What do you think?

evilebottnawi: @ai throw error if env section already define?

ai: @evilebottnawi yeap

evilebottnawi: @ai not throw?

ai: I will fix it (plan to release 3.0 today)

evilebottnawi: @ai :+1:

PR #236, opened 2018-03-15
Proposal: `cover 70% in my stats` query

Vasfed:
When working on projects with existing audiences there're times when you want to cover most of your gathered stats.

This PR adds query `cover 70% in my stats` that returns most popular browsers that cover provided percent of custom stats when combined.


ai:
Good idea!

But if we are adding `cover 70% in my stats`, we also need `cover 70%` and `cover 70% in US`. Do you want to add them to PR?

Vasfed:
@ai already have that, but it exceeds size limit by 44 bytes, now looking into shrinking something

ai:
@Vasfed just increase limit :D

ai:
Awesome. Tomorrow is my open source day. I will accept and release it.

ai:
Released in 3.2

PR #237, opened 2018-03-16
Ability to calculate coverage over custom stats

Vasfed:
Allows to use `mystats` as region name in coverage calculation


ai:
I renamed value to `"my stats"` cea0243

ai:
Released in 3.2

PR #251, opened 2018-04-24
Remove last 2 versions usage from docs

jamiebuilds:


ai:
I keep `last 2 versions`, just highlight the problem part.

I think it will be better to write a big section about queries recommendation at all. I created an issue: https://github.com/browserslist/browserslist/issues/252

I will mention, that using `last 2 version` will select too many browsers and you should think about some `not` selector or `>0.5% in your country`.

Issue #261, opened 2018-05-27
Outdated caniuse-lite warning


ai:

ai:
Done https://github.com/browserslist/browserslist/pull/267/

PR #265, opened 2018-05-31
Outdated caniuse-lite warning #261

ur300:
/cc @ai hi, can you, please, review?

### DIFF
--- a/index.js
+++ b/index.js
@@ -7,6 +7,7 @@ var BrowserslistError = require('./error')
 var env = require('./node') // Will load browser.js in webpack
 
 var FLOAT_RANGE = /^\d+(\.\d+)?(-\d+(\.\d+)?)*$/
+var latestVersion = 0
 
 function normalize (versions) {
   return versions.filter(function (version) {
@@ -154,6 +155,14 @@ function resolve (queries, context) {
   }, [])
 }
 
+function checkLatestVersion (releaseDate) {
+  for (var version in releaseDate) {
+    if (releaseDate[version] > latestVersion) {
+      latestVersion = releaseDate[version]
+    }
+  }
+}
+
 /**
  * Return array of browsers by selection queries.
  *
@@ -279,6 +288,7 @@ browserslist.parseConfig = env.parseConfig
 browserslist.readConfig = env.readConfig
 browserslist.findConfig = env.findConfig
 browserslist.loadConfig = env.loadConfig
+browserslist.testCanIUse = env.testCanIUse
 
 /**
  * Return browsers market coverage.
@@ -752,7 +762,9 @@ var QUERIES = [
         }
       }
     }
+    checkLatestVersion(agents[name].release_date)
   }
+  browserslist.testCanIUse(latestVersion)
 }())
 
 module.exports = browserslist
--- a/node.js
+++ b/node.js
@@ -12,6 +12,7 @@ var FORMAT = 'Browserslist config should be a string or an array ' +
 
 var filenessCache = { }
 var configCache = { }
+var DAYS_TO_UPDATE_CANIUSE = 182
 
 function checkExtend (name) {
   var use = ' Use `dangerousExtend` option to disable.'
@@ -94,6 +95,12 @@ function parsePackage (file) {
   }
 
   return list
+}
+
+function padToEnd (num, digits) {
+  var arr = new Array(digits)
+  arr.unshift(1)
+  return num * arr.join('0')
 }
 
 module.exports = {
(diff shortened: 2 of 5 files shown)


index.js
```diff
@@ -279,6 +288,7 @@ browserslist.parseConfig = env.parseConfig
 browserslist.readConfig = env.readConfig
 browserslist.findConfig = env.findConfig
 browserslist.loadConfig = env.loadConfig
+browserslist.testCanIUse = env.testCanIUse
```
ai: Since this method showing warning, it is better to not put it to public API. How will you use this method? ;)

index.js
```diff
@@ -154,6 +155,14 @@ function resolve (queries, context) {
   }, [])
 }
 
+function checkLatestVersion (releaseDate) {
```
ai: If we show a warning only in Node.js, why do we need this code in non-Node.js Browserslist? ;) Just move everything to `node.js` to keep Browseslist small in browser build.

node.js
```diff
@@ -12,6 +12,7 @@ var FORMAT = 'Browserslist config should be a string or an array ' +
 
 var filenessCache = { }
 var configCache = { }
+var DAYS_TO_UPDATE_CANIUSE = 182
```
ai: `6 * 30` will be more clear

test/caniuse.test.js
```diff
@@ -0,0 +1,13 @@
+var browserslist = require('../')
+
+console.warn = jest.fn()
+
+it('test caniuse version doesn\'t throw warning', () => {
+  expect(browserslist.testCanIUse(1526515200000)).toBeUndefined()
```
ai: You do not test how `checkLatestVersion` work. It is better to change date inside Browserslist (you can find this method in `last n versions` tests)

index.js
ur300: true. i would rather do everything in node.js but i saw a loop iterating through caniuse.agents in index.js so i decided to use this loop and not duplicate it in node.js

test/caniuse.test.js
```diff
@@ -0,0 +1,13 @@
+var browserslist = require('../')
+
+console.warn = jest.fn()
```
ai: This changes should be applied in tests, not in global context

test/caniuse.test.js
```diff
@@ -0,0 +1,13 @@
...
+
+it('test caniuse version doesn\'t throw warning', () => {
+  expect(browserslist.testCanIUse(1526515200000)).toBeUndefined()
+})
+
+browserslist.testCanIUse(1506015200000)
```
ai: Seems like you forget to move this method in some test

node.js
```diff
@@ -258,5 +265,24 @@ module.exports = {
...
+    var now = Date.now()
+    var digitsDiff = String(now).length - String(latestVersion).length
+    if (digitsDiff) {
+      latestVersion = padToEnd(latestVersion, digitsDiff)
+    }
+    var diffDates = now - latestVersion
+    diffDates = diffDates / 1000 / 60 / 60
+    var daysFromUpdate = Math.floor(diffDates / 24)
+    if (daysFromUpdate > DAYS_TO_UPDATE_CANIUSE) {
+      var pathToYarnLock = path.normalize('./yarn.lock')
```
ai: Not correct way to test `yarn.lock`. What if developer call Browserslist in project subdir? Look how we find `package.json`

node.js
```diff
@@ -258,5 +265,24 @@ module.exports = {
   clearCaches: function clearCaches () {
     filenessCache = { }
     configCache = { }
+  },
+
+  testCanIUse: function testCanIUse (latestVersion) {
+    var now = Date.now()
+    var digitsDiff = String(now).length - String(latestVersion).length
```
ai: Why we need this variable?

node.js
```diff
@@ -258,5 +265,24 @@ module.exports = {
...
+    }
+    var diffDates = now - latestVersion
+    diffDates = diffDates / 1000 / 60 / 60
+    var daysFromUpdate = Math.floor(diffDates / 24)
+    if (daysFromUpdate > DAYS_TO_UPDATE_CANIUSE) {
+      var pathToYarnLock = path.normalize('./yarn.lock')
+      var isYarnInstalled = isFile(pathToYarnLock)
+      var packageManagerStr = isYarnInstalled ? 'yarn upgrage' : 'npm update'
+      console.warn('\x1b[31m\x1b[1m[Browserslist] WARN: ' +
+        '\x1b[0m\'caniuse-lite\' is outdated. ' +
```
ai: Now every terminal support color output

node.js
```diff
@@ -258,5 +265,24 @@ module.exports = {
...
+    if (digitsDiff) {
+      latestVersion = padToEnd(latestVersion, digitsDiff)
+    }
+    var diffDates = now - latestVersion
+    diffDates = diffDates / 1000 / 60 / 60
+    var daysFromUpdate = Math.floor(diffDates / 24)
+    if (daysFromUpdate > DAYS_TO_UPDATE_CANIUSE) {
+      var pathToYarnLock = path.normalize('./yarn.lock')
+      var isYarnInstalled = isFile(pathToYarnLock)
+      var packageManagerStr = isYarnInstalled ? 'yarn upgrage' : 'npm update'
```
ai: `npm update` will update many packages and could break an application. A user needs to update only `caniuse-lite` and `browserslist`.

node.js
```diff
@@ -258,5 +265,24 @@ module.exports = {
...
+  },
+
+  testCanIUse: function testCanIUse (latestVersion) {
+    var now = Date.now()
+    var digitsDiff = String(now).length - String(latestVersion).length
+    if (digitsDiff) {
+      latestVersion = padToEnd(latestVersion, digitsDiff)
+    }
+    var diffDates = now - latestVersion
+    diffDates = diffDates / 1000 / 60 / 60
```
ai: It will be better to have `MS_TO_UPDATE_CANIUSE` instead of `DAYS_TO_UPDATE_CANIUSE`, so you will not call calculation every time.

node.js
ur300: caniuse release_date has 10 digits while js Date in milliseconds has 13

node.js
ur300: seems you missed this part of message at the end ```'\x1b[0m\'caniuse-lite\' is outdated.```

node.js
ai: Because `release_date` contains seconds, and `Date.now` milliseconds. You algorithm too complex and will not work in some cases (like 999 in `release_date` and 1000_000 in `Date.now`)

node.js
ai: Yeap. The code is too complex :(. Make it clear by `Run ' + packageManager + ' update caniuse-lite browserslist '`

PR #267, opened 2018-06-02
Added caniuse-lite browsers release date check

An-Tu:


test/release.test.js
```diff
@@ -0,0 +1,105 @@
...
+  })
+
+  afterEach(() => {
+    browserslist.data = originData
+    console.warn = originWarn
+  })
+
+  it('checkCanIUse doesn\'t throw warning', () => {
+    browserslist('last 2 versions')
+    expect(console.warn).toHaveBeenCalledTimes(1)
```
ai: It is better to check exact text (maybe Jest will throw warning too)

node.js
```diff
@@ -258,5 +272,33 @@ module.exports = {
...
     configCache = { }
+  },
+
+  checkCanIUse: function checkCanIUse (agentsObj) {
+    var newestBrowserReleaseTime = getNewestReleaseTime(agentsObj)
+    var today = new Date()
+    var halfYearAgo = new Date(
+      today.getFullYear(),
+      today.getMonth() - MONTH_TO_UPDATE_CANIUSE,
+      today.getDate()
```
ai: It could generate invalid date if today in 31 days of month and `month - 6` will have only 28/30 days.

It is better just use `Date.now() - 6 * 30 * 24 * 60 * 60 * 1000`

node.js
```diff
@@ -258,5 +272,33 @@ module.exports = {
...
+    ).getTime()
+
+    var isWarningNeed = newestBrowserReleaseTime - halfYearAgo < 0
+
+    if (isWarningNeed) {
+      var hasYarnLock = false
+      eachParent(__filename, function (dir) {
+        var yarnLock = path.join(dir, 'yarn.lock')
+        if (isFile(yarnLock)) {
```
ai: It is better to check `package.json` and `yarn.lock` in the same dir

node.js
```diff
@@ -258,5 +272,33 @@ module.exports = {
...
+          hasYarnLock = true
+        }
+      })
+
+      var packageManager = hasYarnLock ? 'yarn upgrade' : 'npm update'
+      console.warn('\x1b[1m!!![Browserslist] WARN: ' +
```
ai: Not every terminal has color support

test/release.test.js
ai: Also, we need to check `yarn/npm` detection. You can do it by mocking `fs` methods.

ai:
Thanks, good PR. But you need a few fixes (I added comments).

ai:
1. You need to fix Travis CI (reduce the size or increase limit).
2. By some reason, all tests print the warning.
3. Warning text will be better without `!!!`. It is not so urgent.

An-Tu:
Is need to remove warning from all tests? Because "years" and "since" tests use releseDate which older than six month.

ai:
Yeap. Right now test output is noisy. You can fix this tests.

An-Tu:
Ok, I got it

An-Tu:
Unfortunately, I don't know how to reduce the limit :(

ai:
Awesome :+1: 

I put your name to the Cult of Martians task, but will merge it and release only after PiterCSS (need more time for slides).

An-Tu:
Hooray, Thak you! It was my first PR to the open-source project.
See you at PiterCSS =)

ai:
Find me on piterCSS. I will give a sockets to you.

gaearon:
This warning is pretty confusing. I encountered it when building a CRA project: https://github.com/facebook/create-react-app/issues/6708.

It is **completely unactionable** to our users who don't even know what `Browserslist` or `caniuse-lite` is. The suggestion to add a new version might actually subtly break their setups (this has happened before). In general it's better if a package doesn't assume it's a top-level dependency, and that the consumer is using it directly.

ai:
> This warning is pretty confusing

Sure. What is your suggestion?

> The suggestion to add a new version might actually subtly break their setups

Yeap, `npm update` update everything, which is a little dangerous.

Do you know how to update only specific npm package deeply in an application?

AndrewSouthpaw:
The suggestion in [facebook/create-react-app#6708](https://github.com/facebook/create-react-app/issues/6708) was to delete that entry out of the `(npm|yarn).lock` file and then re-running the install (`npm install` / `yarn`), and that seemed to fix the issue. Might be enough to suggest as much with this warning.

csimi:
It's especially weird if you're targeting ES5 (as a library or due to company requirements). Absolutely unactionable since updating will do nothing in that case.

PR #268, opened 2018-06-11
Update notification for caniuse-lite

Dubprocessor:
It outputs a message to the terminal notifying the user about the need to update.


ai:
1. You will not be able to call `updateNotifier.caniuseUpdateNotifier` and `updateNotifier` will be `false` in webpack build ;)
2. Can you use more compact methods names in `updateNotifier`. Uglify can’t compress method names.
3. Maybe we should put `update.js` to `node.js`?
4. You lost very useful yarn/npm detection. Can you just copy old message?

Dubprocessor:
1 Maybe i did something wrong in  browser  field of package.json? :)
2 Done.
3 Done.
4. Final decision is up to you, of course. You are great yarn advocate! ;)  I think, that "yarn" detection is not so useful. Many people will be recieving messages for "yarn" even if they haven't  "yarn" installed. Providing instructions for both kind of users is common practice.
What do you mean saying "old message"? The message should be one line?

ai:
Why do you think yarn message will be irrelevant to users with `yarn.lock` in project root?

Dubprocessor:
Actually, I'm one of  PostCSS  users and this message is for me ;)
And i had not `yarn ` installed before i forked this repo. But in many of my projects i have `yarn.lock`  in root dir, because i just do not delete it. It comes from `create-react-app` or other boilerplates and still stay in the root directory without any reason for a long time..I know i should delete it, but i usually don't. I think that there is a lot of disorganized people like me.

ai:
If you have `yarn.lock` but doesn't use Yarn, it is a bad practice. In PostCSS you just don't use lock file and could have different test behavior.

ai:
I manually cherry picked changes 29edf22

Dubprocessor:
You're right, but I always have a lock file in the root directory. It is called package-lock.json. Npm generates it since v5. And here an example when people use both managers at the same time.
https://stackoverflow.com/questions/44552348/should-i-commit-yarn-lock-and-package-lock-json-files

PR #281, opened 2018-08-10
Fixed small typo

joshingmachine:
I believe it should be "duplicate" in those cases. Let me know if there's anything else I should/can do to help.


ai:
Fixed in 4.0.2

Issue #284, opened 2018-08-10
Should "extends @company/browserslist-config/desktop" work?


Schweinepriester:
Hi!

We have published our browserslist-config as `@company/browserslist-config` and are now trying to reference specific files:

```json
"browserslist": [
  "extends @company/browserslist-config/desktop",
  "extends @company/browserslist-config/mobile"
],
```

```bash
[farmer@2c0975f0e531 app]$ npm ls browserslist
…
`-- autoprefixer@9.1.0
  `-- browserslist@4.0.1
```

This fails:

```bash
[11:54:51] 'autoprefixer' errored after 155 ms
[11:54:51] BrowserslistError in plugin "gulp-postcss"
Message:
    Browserslist config needs `browserslist-config-` prefix.  Use `dangerousExtend` option to disable.
Details:
    browserslist: true
    fileName: /app/src/qbShiny.scss
    domainEmitter: [object Object]
    domain: [object Object]
    domainThrown: false

Stack:
BrowserslistError: Browserslist config needs `browserslist-config-` prefix.  Use `dangerousExtend` option to disable.
    at checkExtend (/app/node_modules/browserslist/node.js:21:11)
    at Object.loadQueries (/app/node_modules/browserslist/node.js:116:35)
    at Function.select (/app/node_modules/browserslist/index.js:735:26)
    at /app/node_modules/browserslist/index.js:141:33
    at Array.reduce (<anonymous>)
    at resolve (/app/node_modules/browserslist/index.js:122:18)
    at browserslist (/app/node_modules/browserslist/index.js:219:16)
    at Browsers.parse (/app/node_modules/autoprefixer/lib/browsers.js:65:12)
    at new Browsers (/app/node_modules/autoprefixer/lib/browsers.js:50:26)
    at loadPrefixes (/app/node_modules/autoprefixer/lib/autoprefixer.js:78:20)
```

Without referencing works:

```json
"browserslist": [
    "extends @company/browserslist-config"
  ],
```

==> Should `"extends @company/browserslist-config/desktop"` work?

---

Imho <https://github.com/browserslist/browserslist#shareable-configs> doesn't state that scoped packages and referencing specific files is mutually exclusive.

Thus I'd argue that `SCOPED_CONFIG__PATTERN` at https://github.com/browserslist/browserslist/blob/master/node.js#L9 is too strict and should be something like

```js
[...]

PR #285, opened 2018-08-10
Allow "@scope/browserslist-config/somefile"

Schweinepriester:
See #284.

### DIFF
--- a/node.js
+++ b/node.js
@@ -6,7 +6,7 @@ var BrowserslistError = require('./error')
 
 var IS_SECTION = /^\s*\[(.+)\]\s*$/
 var CONFIG_PATTERN = /^browserslist-config-/
-var SCOPED_CONFIG__PATTERN = /@[^./]+\/browserslist-config(-|$)/
+var SCOPED_CONFIG__PATTERN = /@[^./]+\/browserslist-config(-|$|\/)/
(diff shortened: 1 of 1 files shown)


ai:
Thanks. Right now I am on Def Con. I will review it on next week.

ai:
With this regexp you can pass something `extends @good/browserslist-config/../evil/hack.js`.

Can you make regexp more strict?

ai:
Nope, it is not secure too. Attacker can just write `extends @good/browserslist-config/subdir/../../evil/hack.js` 😉

We need to find `/./` and `/../` in any part. Maybe we should use two RegExp.

Schweinepriester:
Sorry for the late response, but doesn't https://github.com/browserslist/browserslist/blob/master/node.js#L24-L27
```js
if (name.indexOf('.') !== -1) {
  throw new BrowserslistError(
    '`.` not allowed in Browserslist config name. ' + use)
}
```
already cover this, by generally disallowing any `.` in the config name, thus also rejecting `/./` as well as `/../`?

If so, I believe the current changes in the pull request would suffice.

If not, I experimented with https://regex101.com/r/N4SzUq/1 as second RegExp.

ai:
Right, I forgot about `checkExtend`

ai:
Released in 4.3

Issue #283, opened 2018-08-10
Counter intuitive filtering


dotnetCarpenter:
`browserslist` uses a logical `OR` when combining queries which I think is very unfortunate and I did not find any discussion about this in the issue list here.

From the [README Notes](https://github.com/browserslist/browserslist#notes):
> Multiple criteria are combined as a boolean `OR`. A browser version must match at least one of the criteria to be selected.

### Problem
When specifying a set of browsers, it's natural to assume that you create selection/filter of all browsers while constructing your query.

Query example: 
+ `last 2 versions` - should give the last 2 versions of all browsers
+ `last 2 version, not IE > 11` - should give the last 2 versions of all browsers, but no IE below 11
+ `last 2 version, not IE > 11, > 1%` - should give the last 2 versions of all browsers, but no IE below 11 and not any browsers with less than 1% global usage.

**These examples requires that queries are combined with a logical `AND` operator.**

Say you want to target the last version of all browsers, but no IE below 11 and not any browsers with less than 1% global usage. With `browserslist` you **can not**. Well, you can with `> 1%, not IE < 11, not Chrome < 67, not iOS < 11.4` but you have to update the query for each new browser release.

I realize that this is a breaking change and requires a major version bump but I think it's an important discussion and I hope I can convince you all of the usefulness of this.

PS. I came to this issue because I wanted to configure babel to not transpile `const` and `let` to `var` , since I only support browsers that also support `const` and `let`. I'm not sure if it's ultimately is a bug in babel but that I can not target the browsers I want with `browserslist` seems to be the first stop-gap.

Ref: https://github.com/jonrohan/browserl.ist/issues/98

ai:
Let's just add `and` operator to config parser. Do you want to send PR?

dotnetCarpenter:
If there is a will to merge it, I will take a look.

Perhaps it is a good idea to get a release candidate with the PR and then I
can take a look at Babel and see if that is enough to configure the
transpiling of js `let`/`const` statements, which is my current use-case?

My point being, that the implementation have to be useful, end-to-end, for
at least one use-case.

On 10 Aug 2018 16:34, "Andrey Sitnik" <notifications@github.com> wrote:

Let's just add and operator to config parser. Do you want to send PR?

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/browserslist/browserslist/issues/283#issuecomment-412101093>,
or mute the thread

ai:
[...]

PR #286, opened 2018-08-12
Add "and" operator to query parser

dotnetCarpenter:
add and operator to config parser - fixing #283 

- [x] added failing test for AND operator
- [x] added AND operator 
- [x] discuss AND syntax with the NOT filter
- [x] re-write implementation
- [x] write docs
- [x] optimize implementation

@ai I've only added a failing test for the new feature - do you agree that the test is correct? If yes, then I will implement the `AND` operator as defined on [line 297](286/files#diff-f9ecb44bab145f0b1a5b313a1fab57aeR297).

```js
browserslist('last 1 version and not IE < 11 and > 1%')
```

EDIT: fixed wrong test string for IE version


ai:
Test is correct, but fixtures are too long. It is better just to use different queries.

dotnetCarpenter:
Oops - should have been `browserslist('last 1 version and not IE < 11 and > 1%')` - less than IE11

dotnetCarpenter:
@ai Is there not a existing fixture in the fixture folder that I can use?

ai:
Just don't use `> 1%`. `ie >= 6 and ie <= 7` will be enough. Your code will be independent for queries combined by `and`.

dotnetCarpenter:
@ai I've implemented this a little week ago but I got worried about the `AND` syntax in combination with the exclude `NOT` because it's confusing.

It's not the smallest implementation I can think of but I would like to discuss the syntax before I write more optimized code for this.

It's getting late here, but I just wanted to update this PR before I go to bed. I'll write more tomorrow :)

dotnetCarpenter:
The anatomy of a search query (string):

Say we have `ie > 6, not ie 7`, which will give us the set `[ie 11, ie 10, ie 9, ie 8]`.
First we have to parse the query.
OR string
_The first part is an initial query that provides a set that we can run our exclusive query on._

Exclusive queries are not standalone. Exclusive queries only works if there is a set to filter from, so they don't have their own query type. What makes the type, can be either of 3 things. Either it's the initial query (or a oneliner), or it's an OR or AND query. We'll see in a moment what the main difference is.
AND string

In both cases we will first get an array of `["ie 7", "ie 8", "ie 9", "ie 10", "ie 11"]`.
No matter the query type the exclusive query will get a list of `["ie 7"]` and then depending on the query type, it will filter the result using the old behaviour or the new.
The old behaviour is used for `QueryType.oneliner` and `QueryType.or`.
1. First add a zero version to each item in the query array called `array`.
2. Then remove all items in the `result` array with what is in the query array `array`.
3. Return the modified `result` array.
```js
if (isExclude) { // array -> ["ie 7"]
  array = array.concat(array.map(function (j) { // array -> ["ie 7", "ie 0"]
    return j.replace(/\s\S+/, ' 0')
  }))
  return result.filter(function (j) { // result -> ["ie 8", "ie 9", "ie 10", "ie 11"]
    return array.indexOf(j) === -1
  })
}
```

With an AND query things are a little different but the end result is the same.
1. Remove all items in the `result` array with what is in the query array `array`.
2.  Return the modified `result` array.

```js
if (isExclude) { // array -> ["ie 7"]
  return result.filter(function (j) { // result -> ["ie 8", "ie 9", "ie 10", "ie 11"]
    return array.indexOf(j) === -1
  })
}
```
_For `QueryType.and` the filter is a little more straight forward._

### So why is there two different ways to do the same and what is the main difference between `AND` and `OR`?

If we add a third query, like `last 2 versions` to the above we will get vastly different results.
or_versions

With an OR query: `ie > 6, not ie 7 or last 2 version` (in this PR you can write `,`/`or` interchangeably). 

The third query will get a set of `["ie 10", "ie 11", "edge 16", "edge 17", "firefox 60", "firefox 61", "chrome 66", "chrome 67", ...]` and merge it with the `result` array (`return result.concat(array)`).

The end result will be:
`["and_chr 67", "and_ff 60", "and_qq 1.2", "and_uc 11.8", "android 67", "android 4.4.3-4.4.4", "baidu 7.12", "bb 10", "bb 7", "chrome 67", "chrome 66", "edge 17", "edge 16", "firefox 61", "firefox 60", "ie 11", "ie 10", "ie 9", "ie 8", "ie_mob 11", "ie_mob 10", "ios_saf 11.3-11.4", "ios_saf 11.0-11.2", "op_mini all", "op_mob 46", "op_mob 12.1", "opera 53", "opera 52", "safari 11.1", "safari 11", "samsung 7.2", "samsung 6.2"]`
_Notice that the previous filtered out IE versions are back again._
and_versions
If we instead wrote `ie > 6, not ie 7 and last 2 version`, we would get `["ie 11", "ie 10"]. Since the AND query is filtering the `result` array.
```js
case QueryType.and:
  if (isExclude) {
    return result.filter(function (j) {
      return array.indexOf(j) === -1
    })
  } else { // array -> ["ie 10", "ie 11", "edge 16", "edge 17", "firefox 60", "firefox 61", "chrome 66", "chrome 67", ...]
    return result.filter(function (j) { // result -> // ["ie 11", "ie 10"]
      return array.indexOf(j) !== -1
    })
  }
```

A computational faster version of this would be `ie > 6 and not ie 7 and last 2 version`, since the and filter is simpler.

Originally we used to split each query on `,` and then inside each query handling code (in the `resolve` function), we would either make it an exclusive filter query or a OR `concat` query. Since `last 2 versions, not IE < 11 and > 1%` and `> 0.2% and last 2 versions or safari 12 and not safari < 12` are valid queries. Both OR and AND query can be an exclusive query. The way I deal with it is to parse the query string backwards, since it is the separator along with the query text that is significant for which type of query we want to execute. In other words, the separator signifies what to do with the `result` array and the query text signifies if we should filter the `result` array or not.

1. does text has `not` -> yes, then filter `result`
2. is it an OR query -> yes, then `concat` `array` to `result`
3. is it an AND query -> yes, then filter the `result`
_There is a duplication of meaning here._

### This brings me to the query syntax

It's confusing that `not` and `and` does the same thing. The reason to introduce it was to fix OR query but it's just another `not` that works with better natural language.
`ie > 6, not ie 7 and last 2 version` can not be written in a way that takes the last 2 versions of IE without the new AND query but the resulting syntax for `browserslist` is muddy at best.
The fix is `ie >= 5.5 and last 2 versions`.

But `last 2 versions, not IE < 11` and `last 2 versions and not IE < 11` is the exact same! Even if the latter is slightly more efficient, then there is no conceptual difference.

`last 2 versions or not IE < 11` is even worse but while I could remove `or` being the same as `,`, keeping it actually makes `ie > 6, not ie 7 or last 2 version` a lot more comprehensible than `ie > 6, not ie 7, last 2 version` giving a plethora of browsers.

Saying that:

>Multiple criteria are combined as a boolean `OR`.

is indeed wrong!
**It would be correct to say, each query that does not have `not`, is combined with logical `OR`.**
Since a `not` query will always filter on the `result` array and never be `concat`/combined with the `result` array.

I have a pretty good idea of how I would construct the query syntax to make better sense, but not while keeping backwards compatibility. I'm afraid that this PR and what it fixes, just makes the query syntax a lot worse. 

1. **This fix is needed**
2. It would be good to agree on a new and better syntax for version 5.
3. It would be very good indeed, to not call the new filter query for `and` but something closer to `not`.

@ai What are your thoughts?

My current implementation makes it easy to add or change keywords like `,`, `OR` and  `AND`. I also mark a difference between the initial/oneliner query and OR and AND queries, although it is unused now, I think it's useful if we change the syntax. The `not` query is decided just in time after executing the `query.queryString` and is currently not a query type. I have made as few changes to the original code as possible. But if we agree to better syntax, I think we can also remove some code duplication and simplify things a little.

ai:
Wow! The comment even with images :).

I will read in next hour after fixing Autoprefixer issue.

.gitignore
```diff
@@ -3,3 +3,4 @@ npm-debug.log
 yarn-error.log
 
 coverage/
+.vscode
```
ai: The project policy is to keep text editor ignores in local `.git/info/exclude` or in system git config `core.excludesFile`

package.json
```diff
@@ -64,7 +64,7 @@
   "size-limit": [
     {
       "path": "index.js",
-      "limit": "8.511 KB"
+      "limit": "8.739 KB"
```
ai: It is OK to just set `"limit": "8.8 KB"`

query.js
```diff
@@ -0,0 +1,6 @@
+function BrowserslistQuery (type, value) {
```
ai: Do we need a separated file just for the simple class without methods? We can use plain object and reduce project size

index.js
```diff
@@ -4,11 +4,19 @@ var jsEOL = require('node-releases/data/release-schedule/release-schedule.json')
...
 var e2c = require('electron-to-chromium/versions')
 
+var BrowserslistQuery = require('./query')
 var BrowserslistError = require('./error')
 var env = require('./node') // Will load browser.js in webpack
 
 var FLOAT_RANGE = /^\d+(\.\d+)?(-\d+(\.\d+)?)*$/
 
+// Enum values MUST be powers of 2, so combination are safe
+var QueryType = {
```
ai: If you will define `var QUERY_OR = 2; var QUERY_AND = 4;` it will reduce project size ;)

ai:
> in this PR you can write ,/or interchangeably

I like it :+1: Cool that you think about it

ai:
> But last 2 versions, not IE < 11 and last 2 versions and not IE < 11 is the exact same!

I think it is not the big problem. Our language have a lot of way to say the same thing and we have not big problem with it :)

You are good developer if you thought about all this edge cases :+1:

ai:
> I have a pretty good idea of how I would construct the query syntax to make better sense, but not while keeping backwards compatibility

Sorry, we just can’t break compatibility. One of the main reason is that one project can use different tools with different Browserslist inside. And all that Browserslist will use the same config.

If we will change the behaviour, users will kill me in this confused case when even old queries work in the different way :sweat_smile:

ai:
Hope I didn’t miss questions in your comment (sorry, it is hard to write big comments after big party :D)

If I missed something feel free to give me a list of questions

dotnetCarpenter:
I appreciate your feedback @ai. On second thought, it should be possible to write `ie > 6, not ie 7, not last 2 version`, which would be equal to `ie > 6 and not ie 7 and last 2 version`. This untested but if it's true, them wr could substitute the `not` implementation with the `and` implementation. So we would have a more advanced`not`, while reducing the code size and be backwards compatible. Just a quick thought and untested.

dotnetCarpenter:
I'll take another look this weekend.

Encapsulating queries into an object is something I want to stick with regardless. It makes it easy to add new operators to the query syntax and refine existing syntax without rewriting a lot of code.

ai:
Queries AST? I like the idea.

dotnetCarpenter:
@ai Just wondering why does the `not` part here create an intermediate array?
```js
if (isExclude) {
  array = array.concat(array.map(function (j) {
    return j.replace(/\s\S+/, ' 0')
  }))
  return result.filter(function (j) {
    // remove result items that are in array
    return array.indexOf(j) === -1
  })
}
```
https://github.com/browserslist/browserslist/blob/master/index.js#L142-L149

As far as I can see it just create version 0 of each browser.

ai:
> As far as I can see it just create version 0 of each browser.

The problem is in Can I Use inconsistency. Some browsers has separated versions `ie 6`, `ie 5`. “Non-white” browsers like Opera Mini or Chinese browsers has only one record for all versions. Some of them use `X all` pattern, done `X 0` pattern. And my favorite part is that in some data Can I Use uses latest real version `X 5` even it uses `X 0` in other data 😭

dotnetCarpenter:
>And my favorite part is that in some data Can I Use uses latest real version X 5 even it uses X 0 in other data :sob:
Right. So do you have a test for that so I don't introduce errors in my new syntax?

query.js
dotnetCarpenter: Well, I added it because there is 2 places here I currently need the `BrowserslistQuery` so I don't break old code. `function resolve (queries, context) {` can also take a array of strings.

+ https://github.com/browserslist/browserslist/pull/286/files#diff-168726dbe96b3ce427e7fedce31bb0bcR300
+ https://github.com/browserslist/browserslist/pull/286/files#diff-168726dbe96b3ce427e7fedce31bb0bcR139

If I can avoid code duplication (multiple places where I initialize the same object) in my re-write, then I'll remove `BrowserslistQuery`.

dotnetCarpenter:
After reading up on Sets on wikipedia, I've found some neat illustrative images which are useful for the new documentation and a missing query that I call `WITHOUT` (symmetric difference).

Normal Query | Specialized Query
------------ | -------------
UNION or `OR` | RELATIVE COMPLEMENT not `AND NOT` `OR NOT`
INTERSECTION and `AND` | SYMMETRIC DIFFERENCE without `WITHOUT`

I have to make an implementation of `WITHOUT` before I fully understand how it fits in but it seem obvious that we should have it after reading the two pages on [`OR`][1] and [`AND`][2].

1. `,`/`OR` https://en.wikipedia.org/wiki/Union_(set_theory)
2. `AND` https://en.wikipedia.org/wiki/Intersection_(set_theory)
3. `NOT` https://en.wikipedia.org/wiki/Complement_(set_theory)#Relative_complement
4. `WITHOUT` https://en.wikipedia.org/wiki/Symmetric_difference

[1]: https://en.wikipedia.org/wiki/Union_(set_theory)
[2]: https://en.wikipedia.org/wiki/Intersection_(set_theory)
[3]: https://en.wikipedia.org/wiki/Complement_(set_theory)#Relative_complement
[4]: https://en.wikipedia.org/wiki/Symmetric_difference

dotnetCarpenter:
`WITHOUT` can be described as `NOT DUPLICATES` but `WITHOUT` reads better I think.

_last 2 versions not duplicates dead_ : <- also conflicts with the `NOT` operator.

_last 2 versions without dead_

dotnetCarpenter:
Or maybe just stay simple and not implement `WITHOUT`. I'll finish this next weekend. Let me know what you think.

ai:
> Right 😱 So do you have a test for that so I don't introduce errors in my new syntax?

To be honest it is hard to cover all cases. But let's hope, that everything is covered.

Anyway I will review your coffee and if we will have some problems, we will just fix them 😎

ai:
I think we should add queries only if there is a practice reasons. So I am OK to not adding `without` 😄.

I think code size and code complexity is more important than supporting all theoretical set.

But if some users will have use cases, we can add it later.

ai:
Do you need help with it?

dotnetCarpenter:
@ai thanks for asking! To be honest, I haven't even had time to look at this since 23 days ago. I probably won't have time until beginning of next month.

If you want, I'll be happy to review and come with suggestions. Or if you can wait, I'll do it in a couple weeks. I think it's a 4-5 hours job at this point.

ai:
I will wait (anyway I will be free only on next week too :D)

ai:
BTW, I really like this PR :D

ai:
What do we need to fix to finish this PR? I could ask my friends to do it.

dotnetCarpenter:
@ai in short, **time**.

I got one week left on my current project and can prioritize fixing this, right after.

Or you can ask your friends to resolve it. I'll be happy to review.

dotnetCarpenter:
@ai One question, [`resolve`](https://github.com/browserslist/browserslist/pull/286/files#diff-168726dbe96b3ce427e7fedce31bb0bcR137) is called from some `QUERIES`.

[line 782](https://github.com/browserslist/browserslist/pull/286/files#diff-168726dbe96b3ce427e7fedce31bb0bcR782)
[line 815](https://github.com/browserslist/browserslist/pull/286/files#diff-168726dbe96b3ce427e7fedce31bb0bcR815)
[line 828](https://github.com/browserslist/browserslist/pull/286/files#diff-168726dbe96b3ce427e7fedce31bb0bcL749)

Is [`resolve`](https://github.com/browserslist/browserslist/pull/286/files#diff-168726dbe96b3ce427e7fedce31bb0bcR137) strictly internal or do we need to keep the function signature (`object`, `string[]`) as oppose to `BrowserslistQuery`, `string[]`?

dotnetCarpenter:
@ai I've removed the `BrowserslistQuery` class and are now using plain objects. I moved the parse part into `resolve` and made sure that the new `AND` syntax works from a `package.json` file ala:

```json
{
  "browserslist": ["ie 9 and ie 10"]
}
```

The only thing I'm unsure off is if it's ok to keep the simple _exclude_ filter for `AND` queries, since I see that you have updated the `OR` exclude algorithm. I added the same test case as in https://github.com/browserslist/browserslist/commit/0119a369dc146f41aea92649c9947d7bf4311ad8 and are not seeing any issues. If you have an example that could fail with the simple filter I'm using for `AND NOT` queries, I'll be happy to use the same. But I just can not see the issue...

```js
return result.filter(function (j) {
    return array.indexOf(j) === -1
})
```
_Source: https://github.com/browserslist/browserslist/pull/286/files#diff-168726dbe96b3ce427e7fedce31bb0bcR196_

But that's it. I think the rewrite is done and only the documentation stuff is left. What do you think?

ai:
> I've removed the BrowserslistQuery class and are now using plain objects. I moved the parse part into resolve and made sure that the new AND syntax works from a package.json file

LGTM

> I added the same test case as in 0119a36 and are not seeing any issues.

Can you backport this (latest version) algorithm to `not or` https://github.com/browserslist/browserslist/commit/21c25e2c0751107d45f4dc768fecad6d870f4913

dotnetCarpenter:
>Can you backport this (latest version) algorithm to not or 21c25e2

@ai Not sure I get what you are asking. I currently have:

```js
if (isExclude) {
  var filter = { }
  var browsers = { }
  array.forEach(function (j) {
    filter[j] = true
    var browser = j.replace(/\s\S+$/, '')
    browsers[browser] = true
  })
  return result.filter(function (j) {
    return !filter[j]
  })
}
// union of result and array
return result.concat(array)
```
_https://github.com/browserslist/browserslist/pull/286/files#diff-168726dbe96b3ce427e7fedce31bb0bcR204_

For `or not` queries. Isn't that the exact same as in [master](https://github.com/browserslist/browserslist/blob/master/index.js#L165-L177)?

ai:
I am taking about `browsers[browser] = true`. Seems like you do not need it and it migrated to the code from old version of my `not or` algorithm.

dotnetCarpenter:
Ahh ok. Sure :)

dotnetCarpenter:
@ai quick question. Since browserslist only support node 6+ which has very good ES2015 support, why not upgrade the syntax?

dotnetCarpenter:
I'll start writing the new documentation and clean up comments.

dotnetCarpenter:
@ai Can I assume that the next version with `and` queries will be `4.4.0`, for the documentation?

ai:
Yeap. Minor release.

dotnetCarpenter:
I think it is easier to view the new documentation here: https://github.com/dotnetCarpenter/browserslist/tree/fix_283#readme

What do you think @ai?

I can not think of anything else and in my unscientific benchmarks I see no difference. `time benchmark/run.js ` is the same on the `master` branch and `fix_283` branch.

ai:
> Since version 4.4.0

I prefer to not write versions since this information will become irrelevant very soon when everyone updated Browserslist.

Of course, it is an open question, but we should keep consistency in docs.

> Since version 4.4.0, we support and queries that intersect much like not do but in a more obvious language. For backwards comparability queries as arrays or delimited with an , is considered or queries. and queries must combine queries in the same string. The above can also be rewritten as:

The first section of the docs for very important information. Let’s keep it free from details. You can add `and` to examples.

> Query combiner type | Illustration | Example

I like this table =^_^=

Let’s put it (and text description) to separated `Queries` subsection with own title.

> Notes

You should remove `or` note from that section

dotnetCarpenter:
I've updated the README, CHANGELOG and dependencies (`yarn.lock`).

Not much more to do as far as I can see. **Ready for `publish`?** :tada:

dotnetCarpenter:
https://github.com/browserslist/browserslist/issues/315#issue-373776468 can now be written as `last 2 versions and > 1%`

dotnetCarpenter:
https://github.com/jonrohan/browserl.ist/issues/98#issue-346139959 can now be written as `last 1 version and not IE < 11 and > 0.2%`

> `last 1 version, not IE < 11, > 0.2%`should not show browsers with less than 0.2% but QQ Browser1.2, Android Browser67, Baidu Browser7.12 are all 0% usage browsers that do show up in the list.

The new syntax makes it much easier to apply common logic to query output.

dotnetCarpenter:
My colleague just suggested that `not` standalone should be a valid combiner.

So that `ie > 5.5 not ie 8` should return ie 11, ie 10, ie 9, ie 7, ie 6.

I agree and do not see any backwards compatibility issue. I will have to change the matcher in the parser to look for `not` and not only `and` and `,`/`or`.

If there is the string is `ie > 5.5, not ie 8` then the `,` will be thrown away.
The same goes for:
 `ie > 5.5 or not ie 8` - `or` is thrown away
 `ie > 5.5 and not ie 8` - `and` is thrown away

https://github.com/browserslist/browserslist/pull/286/files#diff-168726dbe96b3ce427e7fedce31bb0bcR161 will also have to change...

ai:
> So that ie > 5.5 not ie 8 should return ie 11, ie 10, ie 9, ie 7, ie 6.

:+1:

ai:
> Not much more to do as far as I can see. Ready for publish? tada

Yeap, everything is OK. Let’s just wait for comments from Twitter (and `X not Y` query support if you want)

dotnetCarpenter:
I began with `X not Y` and realised that you can now write `not ie 6`.. Should that give every browser minus ie6? Do we have a _every browser_  selector?

ai:
> Do we have a every browser selector?

Nope, we do not have it for an important reason to prevent the bad practice.

It is a popular idea to select `all expect IE`, but there is no real use case for it.

This is why `no ie 6` should be considered still as an error.

ai:
> There is just too many cases where regex fails.

OK, let’s do it later. After releasing this PR. 

I will release it tomorrow or day later.

dotnetCarpenter:
@ai Got it working! 🎉  But needs a lot of refactoring. I'll create a new PR in a few days

ai:
Can you remove `CHANGELOG.md` fixes? I prefer to change it in release commit.

ai:
Released in 4.4. Thanks for your awesome work.

PR #287, opened 2018-08-12
Update node-releases

chicoxyzzy:
Also Node 11 release date was added to schedule so we need to filter out it from list of maintained versions by checking that `Date.now()` is bigger than (future) release timestamp.


ai:
I will release it few hours later

ai:
Fixed in 4.0.2

PR #288, opened 2018-08-18
Added Hall-of-fame to README

sergey48k:
Hall-of-fame is a widget that helps you recognize new/trending/top contributors in your repo. It is designed to be very visual, and change frequently. It autoupdates every hour and requires no maintenance.

Note that by accepting this PR, you will make me a contributor, which means my face will be featured as ’new’ on your Hall-of-fame for one week.

Also note that this PR relies on my GitHub account to update stats (sergey48k). I have no problem with that so you can leave it as is. But if you prefer your own GitHub account, sign in to sourcerer.io using your GitHub, and then send the username to me. I will adjust PR accordingly. Hall-of-fame needs 10-15 requests to GitHub API per hour (GitHub rate limit is 5,000 per hour).

I apologize for a ‘cold call’ type of PR, and I am hoping it was not much of an inconvenience.

### DIFF
--- a/README.md
+++ b/README.md
@@ -528,3 +528,14 @@ browserslist.clearCaches();
 
 To disable the caching altogether, set the `BROWSERSLIST_DISABLE_CACHE`
 environment variable.
+
+## Contributors
+
+[![0](https://sourcerer.io/fame/sergey48k/browserslist/browserslist/images/0)](https://sourcerer.io/fame/sergey48k/browserslist/browserslist/links/0)
+[![1](https://sourcerer.io/fame/sergey48k/browserslist/browserslist/images/1)](https://sourcerer.io/fame/sergey48k/browserslist/browserslist/links/1)
+[![2](https://sourcerer.io/fame/sergey48k/browserslist/browserslist/images/2)](https://sourcerer.io/fame/sergey48k/browserslist/browserslist/links/2)
+[![3](https://sourcerer.io/fame/sergey48k/browserslist/browserslist/images/3)](https://sourcerer.io/fame/sergey48k/browserslist/browserslist/links/3)
+[![4](https://sourcerer.io/fame/sergey48k/browserslist/browserslist/images/4)](https://sourcerer.io/fame/sergey48k/browserslist/browserslist/links/4)
+[![5](https://sourcerer.io/fame/sergey48k/browserslist/browserslist/images/5)](https://sourcerer.io/fame/sergey48k/browserslist/browserslist/links/5)
+[![6](https://sourcerer.io/fame/sergey48k/browserslist/browserslist/images/6)](https://sourcerer.io/fame/sergey48k/browserslist/browserslist/links/6)
+[![7](https://sourcerer.io/fame/sergey48k/browserslist/browserslist/images/7)](https://sourcerer.io/fame/sergey48k/browserslist/browserslist/links/7)


sergey48k:
Hi @ai, (a very cool username!)

Thank you for your response.

To your questions:

Design: we are planning to revisit it with our designer, but if you have specific ideas that you'd like to see implemented I am open to hear them.

1. Numbers. For 'new' and 'trending' contributors, the number is the number of commits in the last 7 days. For top contributors, the number is all time commits. The idea here is to have a widget that changes often, but at the same time is not empty if there is no action in a repo. The widget is recomputed every hour.

Check out these guys, for instance https://github.com/koreader/koreader . They show me as new because they accepted my PR, they show 4 guys as trending because they have been actively working in the repo in the last 7 days, and the last two positions are taken by top contributors, to avoid empty space. 

2. New always goes first, Trending follows, Top is the last. The motivation is to let new contributors see their faces for a few days on your README, as a way of saying thank you. Inside of a category, contributors are sorted by commits. The slot allocation works as follows: Up to three slots for new, up to four slots for trending, and the rest is filled with top.

3. Yes, of course. You will need to switch this from pure markdown to HTML embedded to markdown. You should set height and not width, because height is always the same. For instance:



4. Yes, it's not a problem at all.

Thank you for the feedback! Hall-of-fame is brand new, and we are very eager to find ways to improve and distribute the widget. I appreciate your time.

If you are interesting in merging this, please let me know what you want to do with the height. I will be happy to update the PR with adjusted height and having alt removed.

ai:
Thanks for the answer. The latest question. Why I am in the tending. Maybe if user is part of top, we should remove they from trending to make b trending open for other contributors?

sergey48k:
As it is implemented currently, top does not take space from trending. Top simply fills the space that was left unoccupied by new and trending. If we follow your suggestion, we will have fewer trending. For your repo, for instance, there is one new and one trending (you). If we reclassify you as top, there will be no trending left. If there were, say, two trending contributors in browserslist, you would see them both, at the expense of top.

I wonder though if trending people would also want to be seen as top somehow. Maybe something extra on a badge? Might be making it too complicated though.

ai:
My concern about trending, that I will was steal tending place as a maintainer and it should be better to give it to somebody else.

Maybe we can do something in the calculation core. Maybe some extra check. In the case b of Browserslist I am definitely not a trending some I was here from the beginning.

sergey48k:
What I can certainly do is to remove top3 or so from trending if there are
other candidates for trending. This seems easy and reasonable. But this
won't change anything in browserslist today because only two people made
changes in the repo during the last week.

I also sense that maybe you prefer to be in 'top' and not in 'trending'
regardless of the spot availability. If I always put top3 to top and not
trending, the widget will become more static for many repos. I would like
to hear from more people that it's what they prefer.

I will go ahead and implement that extra check for when there are more
trending than 4. But I also understand your hesitation, which I cannot
resolve until I have more opinions. Feel free to close the PR for now, and
we can always reconnect later.

Again, appreciate your feedback and your time!

Best,
Sergey

sergey48k:
FYI, implemented the change to make space for trending at the expense of
top if there are too many people in trending:
https://github.com/sourcerer-io/hall-of-fame/commit/bb87b53fb4bf8354a6ec5dc94b7ae2347911a2d7

Not in prod yet, will go to prod early next week.

S.

sergey48k:
thanks!
S.

PR #294, opened 2018-09-10
Moving package.json query to the top of the queries list

dimitarnestorov:
Since it's recommended it should be at the top. Even though it's bold not every one will read the `We recommend this way.` note.


ai:
Sure, good idea. Can you fix numbers of other list items?

Issue #297, opened 2018-09-13
CLI Request: Add "--all" option


bdkjones:
I'd like to be able to run a command like this:

    browserslist --all ">0.5%, last 2 versions, not dead"

And have browserslist return BOTH the list of supported browsers AND the coverage statistics for the US and world, all in one fell swoop. Right now, to get this information, I must run three separate commands:

    browserslist ">0.5%, last 2 versions, not dead"
    browserslist --coverage ">0.5%, last 2 versions, not dead"
    browserslist --coverage=us ">0.5%, last 2 versions, not dead"

I envision that the coverage information could simply be added to the bottom of the list of browsers. To maintain the standard output format:

    [browser_name] [space] [version string]

I would suggest formatting the last two "coverage" lines as:

    coverage_world [space] 90.34
    coverage_us [space] 87.65

This drops the extra text returned by `--coverage` and allows consumers of the output to still parse it as a newline-separated list of space-separated key-value pairs.

bdkjones:
This proposal would also not break existing consumers of browserlist's output, because the extra two lines would not appear unless the `--all` flag is passed.

ai:
Interesting. Can you tell me about your use case?

I am thinking about the more clear name of the argument. And I do not like to [lock some specific country](https://github.com/browserslist/browserslist/pull/298/files#diff-2cce40143051e25f811b56c79d619bf5R98).

bdkjones:
I have a GUI app where I want to display both the supported browser list AND the overall coverage stats based on the user’s chosen setting. (Similar to http://browsersl.ist ) This app is not in a JavaScript environment, so I can’t use the api directly; I need to go through the command line.

I’d like to get all the info in one shot because that lets me update the UI in real-time as the user types their browser string. Otherwise, each update would require three separate runs of Browserslist.

I had considered a comma-separated list of country codes, such as:

    --all=us,uk,cn

Which would return that info as well as the global coverage. An alternative flag name might be --include-coverage, or similar, but that was close to the existing API, which I didn’t want to change.

bdkjones:
I have a GUI app where I want to display both the supported browser list AND the overall coverage stats based on the user’s chosen setting. (Similar to http://browsersl.ist ) This app is not in a JavaScript environment, so I can’t use the api directly; I need to go through the command line.

I’d like to get all the info in one shot because that lets me update the UI in real-time as the user types their browser string. Otherwise, each update would require three separate runs of Browserslist. 

[...]

PR #298, opened 2018-09-13
--all CLI Option

bdkjones:
As proposed in #297, this implements an `--all` flag on the CLI to return both the supported browsers and US/Global coverage values.

Implemented in a non-breaking way so that users who do not pass `--all` will still get the same output they did before.


bdkjones:
Closed in favor of implementing a `--json` option.

PR #307, opened 2018-10-17
Extend query keywords with new keyword to match all browser versions

vchyzhevskyi:
Hello there,

Recently I was forced to use the browserslist configuration that excludes some browsers from all browsers versions instead to include some browser versions from the list of all browsers versions.

I read the documentation and didn't found the solution for this, so this PR includes the extension described above. Now we can do something like this:

```
all
not ie <= 9
not ios <= 8
```

The configuration above will exclude all versions of IE before the IE 10 and all versions of IOS before the IOS 9.

Please let me know if this could be merged into the main browserslist repo and if there any other changes required for this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -124,6 +124,7 @@ from one of this sources:
 
 You can specify the browser and Node.js versions by queries (case insensitive):
 
+* `all`: match all browsers versions.
 * `> 5%`: browsers versions selected by global usage statistics.
   `>=`, `<` and `<=` work too.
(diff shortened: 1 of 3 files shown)


ai:
`all` is a bad practice. If we will add a keyword, it will promote this bad practice.

There are a few rare cases when it is OK. In this case, we can use ugly `> 0%`.

ai:
Sorry, I know it is hard when somebody rejects your work. But it will the safe Internet.

We must keep silence about this way :D

PR #309, opened 2018-10-23
Update index.js

yukulele:
allow rational year


ai:
@yukulele Thanks. You need also to improve the test.

Issue #311, opened 2018-10-23
'maintained node versions' referring pennding Node version


y-ogura:
Nuxt.js build error, because of 'maintained node versions' referring pennding Node version.

show this error message.
```
Module build failed (from ./node_modules/nuxt/node_modules/postcss-loader/src/index.js):
BrowserslistError: Unknown version 11 of Node.js
```
![2018-10-23 12 50 15](https://user-images.githubusercontent.com/30112982/47334583-5271f680-d6c2-11e8-8ae6-17fb85cb3e02.png)

sibiraj-s:
<img width="1411" alt="screenshot 2018-10-23 at 11 02 16 am" src="https://user-images.githubusercontent.com/20282546/47337795-200ccd00-d6b3-11e8-9a38-77735d79c022.png">

dangreen:
All my CI builds fall with this error.

CosticaPuntaru:
a ugly hack to allow you to work on localhost:

node_modules/browserslist/index.js:678
![image](https://user-images.githubusercontent.com/6418405/47344064-39f8e080-d6b0-11e8-8338-505be0b24498.png)

andreasradloff:
Changing `maintained node versions` to `node 10` in your .browserslistrc works as a temp fix.

ai:
Oh. I waked up and need 15 minutes for breakfast. Then I will investigate the source of the error and how to fix it.

dangreen:
@ai https://github.com/browserslist/browserslist/pull/312

ai:
Fixed in 4.3.2

PR #312, opened 2018-10-23
Possible resolution of #311

dangreen:
https://github.com/browserslist/browserslist/issues/311#issuecomment-432126592

> Seems browserslist firstly get all versions by user's query (maintained node versions), then checks this versions separately:
> 
> maintained node versions -> ..., v10, v11 (https://github.com/browserslist/browserslist/blob/master/index.js#L706, https://raw.githubusercontent.com/nodejs/Release/master/schedule.json, v11 is exist)
> node 10 -> OK
> node 11 -> ERROR (https://github.com/browserslist/browserslist/blob/master/index.js#L679, https://nodejs.org/dist/index.json, v11 - doesn't exist)

Solution: check EOL version is in released versions.

### DIFF
--- a/index.js
+++ b/index.js
@@ -9,6 +9,16 @@ var env = require('./node') // Will load browser.js in webpack
 
 var FLOAT_RANGE = /^\d+(\.\d+)?(-\d+(\.\d+)?)*$/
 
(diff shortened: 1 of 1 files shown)


ai:
Fixed in 4.3.2

Issue #328, opened 2018-12-05
Internal: Replace electron-to-chromium with electron-release


mutech:
Following a discussion in #326, I'm proposing to replace electron-to-chromium with electron-release.

Advantages:
* Maintained by electron team
* More information (e.g. Node version)

ai:
Great idea. Do you want to send PR?

Kilian:
Hi, I'm both the maintainer of `electron-to-chromium` and part of the electron governance team. 

I can add a node-versions.js to `electron-to-chromium` for you to consume, that'll be substantially lighter than using `electron-releases`. I got this request before: https://github.com/Kilian/electron-to-chromium/issues/17 but I wasn't sure about its use cases. 

I'm still not sure, because the node and the chrome part of a given electron release share their V8 and so their javascript support with the only difference being internal API's, but if it's useful for browserslist consumers then that shouldn't be too hard to add.

ai:
We don’t need Node.js version right now

Issue #329, opened 2018-12-05
Add `node X` to Electron query output


mutech:
See #326

We should check how that changes the caniuse feature selection though.

ai:
Do you want to send PR?

PR #331, opened 2018-12-06
Electron resolves to node+chrome, electron-releases

mutech:
Electron queries resolve to both chrome and node versions, replaced electron-2-chromium with electron-releases, major and minor versions resolve to oldest concrete version (3 -> 3.0.0, 1.1 -> 1.1.0).

Uses electron-releases as devDependency and generates e2cn.js via `npm run importdata` to reducce the package size penalty.

This still breaks the size limit of 9K (10.7K), but that seems to be an acceptable margin.

I don't like the solution using a script to extract the data from electron-releases, but I also don't have a good idea how to do it otherwise.


mutech:
Would fix #328 and #329

mutech:
Size is down to 9.8K with the latest commit, but that's about what can be done.

The remaining problem is that the script has to be run to update the packed file whenever the dependency to electron-releases is updated.

ai:
> Uses electron-releases as devDependency and generates e2cn.js via npm run importdata to reducce the package size penalty.

Sorry, I can't manually call `yarn importdata` on every Electron release. It will require a lot of work and I will not able to do it every day.

ai:
There is another way. We can write to Electron developers and soonest to add this “lite” data to their npm package.

ai:
I am closing this PR since manually updates is not a option

PR #352, opened 2019-02-19
Accept string as config property, fixes #347

### LINKED ISSUE #347: Iterating over strings using for..in breaks browser query type checking
(This is a bit of an odd ticket, because it's caused by global pollution of the string object by other modules. I'm filing it here as well in the other modules that are causing issues.)

**Expected Behaviour**
A config of `{ browserslist: 'Chrome > 30' }` is passed to browserslist and works properly.

**Observed Behaviour**
Using the same config, an error is thrown:
```
ERROR in ./index.js
Module build failed (from /home/Repos/node_modules/babel-loader/lib/index.js):
BrowserslistError: [BABEL] /home/Repos/src/index.js: Browserslist config should be a string or an array of strings with browser queries (While processing: "/home/Repos/node_modules/@babel/preset-env/lib/index.js")
  at check (/home/Repos/node_modules/browserslist/node.js:61:11)
  at parsePackage (/home/Repos/node_modules/browserslist/node.js:94:5)
```

When checking the browser query string passed in, a `for..in` is used to iterate over the string ([here](https://github.com/browserslist/browserslist/blob/master/node.js#L93)) which will potentially pull in all kinds of weird globals. A number of the libraries we use in our buildchain add various things to the String prototype which are then passed to the `check` which throws a `FORMAT` error.

A potential fix for this could be to either hoist the iterating over the array into the `parsePackage` function, or wrap the `check` call with a `hasOwnProperty` test.

dmarkhas:
Should support the config as described here:
https://babeljs.io/docs/en/babel-preset-env

### DIFF
--- a/node.js
+++ b/node.js
@@ -88,9 +88,8 @@ function parsePackage (file) {
   var list = config.browserslist
   if (Array.isArray(list)) {
     list = { defaults: list }
-  }
-  
-  Object.keys(list).forEach(function(el) {
+  }  
+  Object.keys(list).forEach(function (el) {
     check(list[el])
   })
 
--- a/node.js
+++ b/node.js
@@ -88,7 +88,7 @@ function parsePackage (file) {
   var list = config.browserslist
   if (Array.isArray(list)) {
     list = { defaults: list }
-  }  
+  }
   Object.keys(list).forEach(function (el) {
     check(list[el])
   })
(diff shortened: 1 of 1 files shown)


ai:
You need to fix tests and add new test

ai:
Thanks. Released in 4.4.2.

PR #371, opened 2019-05-11
Alias mobile browsers to the desktop version

### LINKED ISSUE #367: Wrong chromeandroid version parsing
We found this bug in Babel's test suite, which a few days ago suddenly started failing.

**Current behavior:**

```js
> browserslist("chrome 53")
[ 'chrome 53' ]
> browserslist("chromeandroid 53")
[ 'and_chr 73' ]
```

**Expected behavior:**

```js
> browserslist("chromeandroid 53")
[ 'and_chr 53' ]
```

---

**UPDATE**

It looks like that the problem is that many browsers only list the last version :thinking: 

Schermata da 2019-04-26 11-25-41

nicolo-ribaudo:
Fixes #367

In this PR, browserslist fallback to the desktop version only if the query doesn't match a mobile version. I didn't choose to always fallback in case a mobile browser has a release which doesn't match its desktop counterpart (is it even possible?).

### DIFF
--- a/test/direct.test.js
+++ b/test/direct.test.js
@@ -53,5 +53,16 @@ it('supports Can I Use cutted versions', () => {
 })
 
 it('supports Can I Use missing mobile versions', () => {
-  expect(browserslist('chromeandroid 53')).toEqual(['and_chr 53'])
+  var opts = { mobileToDesktop: true }
+  expect(browserslist('chromeandroid 53', opts)).toEqual(['and_chr 53'])
+  expect(browserslist('and_ff 60', opts)).toEqual(['and_ff 60'])
+  expect(browserslist('ie_mob 9', opts)).toEqual(['ie_mob 9'])
+  expect(browserslist('op_mob 30', opts)).toEqual(['op_mob 30'])
+})
+
+it('missing mobile versions are not aliased by default', () => {
+  expect(browserslist('chromeandroid 53')).not.toEqual(['and_chr 53'])
+  expect(browserslist('and_ff 60')).not.toEqual(['and_ff 60'])
+  expect(() => browserslist('ie_mob 9')).toThrow()
+  expect(() => browserslist('op_mob 30')).toThrow()
 })
--- a/test/direct.test.js
+++ b/test/direct.test.js
@@ -49,5 +49,9 @@ it('supports Safari TP', () => {
 })
 
 it('supports Can I Use cutted versions', () => {
-  expect(browserslist('and_chr 55')).toHaveLength(1)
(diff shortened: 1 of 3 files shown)


test/direct.test.js
```diff
@@ -49,5 +49,9 @@ it('supports Safari TP', () => {
 })
 
 it('supports Can I Use cutted versions', () => {
-  expect(browserslist('and_chr 55')).toHaveLength(1)
+  expect(browserslist('and_uc 10')).toHaveLength(1)
+})
+
+it('supports Can I Use missing mobile versions', () => {
+  expect(browserslist('chromeandroid 53')).toEqual(['and_chr 53'])
```
ai: We need to test all 4 shortcuts

ai:
We need to do it only on a special option.

I can’t release major version (because many old projects are using Browserslist and they will not update it for years). As result, we can’t break API.

test/direct.test.js
```diff
@@ -49,5 +49,20 @@ it('supports Safari TP', () => {
...
+  expect(browserslist('and_ff 60', opts)).toEqual(['and_ff 60'])
+  expect(browserslist('ie_mob 9', opts)).toEqual(['ie_mob 9'])
+  expect(browserslist('op_mob 30', opts)).toEqual(['op_mob 30'])
+})
+
+it('missing mobile versions are not aliased by default', () => {
+  expect(browserslist('chromeandroid 53')).not.toEqual(['and_chr 53'])
+  expect(browserslist('and_ff 60')).not.toEqual(['and_ff 60'])
+  expect(() => browserslist('ie_mob 9')).toThrow()
+  expect(() => browserslist('op_mob 30')).toThrow()
```
nicolo-ribaudo: I don't know why some of these browsers throw and some resolve to another version, but it's the current behavior:
- IE throws: https://browserl.ist/?q=ie_mob+9
- Chrome resolves to a wrong versoin: https://browserl.ist/?q=and_chr+53

ai: `browserl.ist` is not official, it uses old Browserslist. Doesn’t use it in investigation.

ai: Yeap, error is throwing because Can I Use doesn't have info

ai:
Can you add additional option here https://github.com/browserslist/browserslist#js-api

nicolo-ribaudo:
Thanks!

ai:
@nicolo-ribaudo do you have a Twitter account to mention you in release tweet?

ai:
Released in 4.6

nicolo-ribaudo:
Yup, it's @NicoloRibaudo

PR #375, opened 2019-05-20
Match desktop Chrome for evergreen Android WebView versions

### LINKED ISSUE #374: "last 2 versions" (and "defaults") should not include "android 4.4.3-4.4.4"
This technically falls out from how `caniuse` is representing versions for Android, but I think is worth discussing here since it impacts the `defaults` query. 

Relevant data from `caniuse`:
```json
"android": {
    "browser": "Android Browser",
    "abbr": "And.",
    "prefix": "webkit",
    "type": "mobile",
    "usage_global": {
        "2.1": 0.00480563,
        "2.2": 0.00823823,
        "2.3": 0.00480563,
        "3": 0.000686519,
        "4": 0.0988588,
        "4.1": 0.0494294,
        "4.2-4.3": 0.214194,
        "4.4": 0,
        "4.4.3-4.4.4": 0.300009,
        "67": 0
    }
```

The `caniuse` data is only providing one version covering the time since the [Android webview went evergreen in Android 5](https://developer.chrome.com/multidevice/webview/overview), even though there have been multiple versions in between, just like Chrome.

This causes `browserslist` to include `4.4.3-4.4.4` in the `last 2 versions` (and by association `defaults`) query even though it doesn't truly meet the criteria - it just appears to due to how the data is represented.

antross:
Fix #374 

I did some digging into the MDN [browser-compat-data](https://github.com/mdn/browser-compat-data) for a second opinion on version history for the Android WebView. Based on this it appears a more complete way to account for missing version numbers in the `caniuse` data (short of using the MDN data directly) would be to match the version numbers after the Android WebView went evergreen (version 37) with desktop Chrome. 

https://github.com/mdn/browser-compat-data/blob/master/browsers/webview_android.json
https://github.com/mdn/browser-compat-data/blob/master/browsers/chrome.json

This gets `last N versions` and related queries working correctly and resolves another issue with `caniuse` data currently being about seven versions behind the current version of the Android WebView.

This reliable of a relationship between versions doesn't necessarily apply to all mobile browsers, so I left the recently added `mobileToDesktop` mapping untouched.

### DIFF
--- a/index.js
+++ b/index.js
@@ -966,19 +966,40 @@ var QUERIES = [
 // Get and convert Can I Use data
 
 (function () {
+  function getVersions (name) {
+    if (name !== 'android') {
+      return agents[name].versions
+    } else {
+      // Match desktop Chrome for evergreen Android WebView versions
+      var firstEvergreenVersion = 37
+      var beforeEvergreen = agents[name].versions.filter(function (v) {
+        return parseInt(v) < firstEvergreenVersion
+      })
+      var afterEvergreen = agents.chrome.versions
+        .slice(0, -3)
+        .filter(function (v) {
+          return parseInt(v) >= firstEvergreenVersion
+        })
+      return beforeEvergreen
+        .concat(afterEvergreen)
+        .concat([null, null, null])
+    }
+  }
+
   for (var name in agents) {
     var browser = agents[name]
+    var versions = getVersions(name)
     browserslist.data[name] = {
       name: name,
-      versions: normalize(agents[name].versions),
-      released: normalize(agents[name].versions.slice(0, -3)),
+      versions: normalize(versions),
+      released: normalize(versions.slice(0, -3)),
       releaseDate: agents[name].release_date
     }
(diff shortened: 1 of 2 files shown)


antross:
Looking into the test failures. Thought I had run this locally, but apparently not with my changes... 😞

ai:
1. Check that you commit all changes( maybe you forget to commit a file)
2. Try to run `yarn test` instead of `npx jest`, etc

I am preparing to conference talk right now. I will investigate this problem a few days later.

antross:
Turns out I just neglected to re-run the tests after my last edit which also matched `and_chr` 🤦‍♂️

I fixed it by removing that edit and re-scoping this PR to just match data for the Android WebView since that was the original focus. I also added a test case to validate the version transition from "before evergreen" (up to `android 4.4.3-4.4.4`) and "after evergreen" (`android 37` and beyond).

ai:
Oops. I missed one big problem https://github.com/browserslist/browserslist/issues/374#issuecomment-495583421. I reverted PR manually.

PR #378, opened 2019-05-27
implement support for patch versions of Electron

Kilian:
This commit silently changes electron versions with patch, like "4.0.4", to versions without patch ("4.0") so they work with the `electron-to-chromium` version.js file. 

This means people can write their full electron version (or use that automatically) with browserslist.

### DIFF
--- a/index.js
+++ b/index.js
@@ -33,6 +33,17 @@ function normalize (versions) {
   })
 }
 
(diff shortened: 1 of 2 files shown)


ai:
Thanks. Released in 4.6.1

Issue #382, opened 2019-06-18
BrowserslistError: Unknown browser query `android all`. Maybe you are using old Browserslist or made typo in query.


akshayakrsh:
```BrowserslistError: Unknown browser query `android all`. Maybe you are using old Browserslist or made typo in query.```

I received this error when using the `react-starter-app`, maybe its got to do something with `BrowsersList` itself, hence reporting it here.

**Comment by silverwind** [comment]
Same error, my `browserslist` does not actually contain a `android all`, where does it come from?

**Comment by SathyaThangam** [comment]
BrowserslistError: Unknown browser query `android all`. Maybe you are using old Browserslist or made typo in query.
    at Array.reduce (<anonymous>)
    at Array.some (<anonymous>)
    at Array.filter (<anonymous>)
    at new Promise 

> (<anonymous>)

even I received the same error during deployment

**Comment by lkiarest** [comment]
Unknown browser query `android all`. Maybe you are using old Browserslist or made typo in query.

**Comment by zhipenglin** [comment]
Same issue

**Comment by TonSDe** [comment]
I have same issue. and can't run "yarn dev" or "yarn run build" in Nuxt-app.

**Comment by Desyon** [comment]
Having the same issue

**Comment by solonik1** [comment]
Same issue

**Comment by meliksu** [comment]
Same issue

**Comment by iwangx** [comment]
Same issue

**Comment by micooz** [comment]
Same issue :)

**Comment by NoureldeanSaed** [comment]
Same issue

**Comment by pawelrol** [comment]
Same issue

**Comment by hileix** [comment]
Same issue.

**Comment by rajjejosefsson** [comment]
have the issue when using postcss-preset-env using rollup, seems create-react-app uses this too

**Comment by keyurbhole** [comment]
Same Issue

**Comment by ranabhat** [comment]
Same issue 
```
{ BrowserslistError: Unknown browser query `android all`. Maybe you are using old Browserslist or made typo in query.
    at unknownQuery (/home/OCPP-tester-ElifTech/ocpp-tester-masterv2/ocpp-tester/node_modules/browserslist/index.js:204:10)
    at /home/OCPP-tester-ElifTech/ocpp-tester-masterv2/ocpp-tester/node_modules/browserslist/index.js:296:11
    at Array.reduce (<anonymous>)
    at resolve (/home/OCPP-tester-ElifTech/ocpp-tester-masterv2/ocpp-tester/node_modules/browserslist/index.js:237:18)
[...]

PR #384, opened 2019-06-18
Lock caniuse-lite version for temporary bug fix

nmonin-spro:
[caniuse-lite](url) has updated recently to 1.0.30000975 and this issue appeared https://github.com/browserslist/browserslist/issues/382.
Using 1.0.30000974 instead would fix this issue temparary


ai:
We should find who send the wrong query to Browserslist. Not locking users from fresh Can I Use data.

PR #418, opened 2019-10-20
add core-js to the list of Browserslist users

zloirock:
Browserslist config can be used in `core-js-compat` (used by many tools like `@babel/preset-env`) and `core-js-builder`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,6 +8,7 @@ front-end tools. It is used in:
 
 * [Autoprefixer]
 * [Babel]
+* [core-js]
 * [postcss-preset-env]
 * [eslint-plugin-compat]
 * [stylelint-no-unsupported-browser-features]
@@ -64,6 +65,7 @@ Browserslist will take queries from tool option,
 [Can I Use]:                                 https://caniuse.com/
 [Babel]:                                     https://github.com/babel/babel/tree/master/packages/babel-preset-env
 [obsolete-webpack-plugin]:                   https://github.com/ElemeFE/obsolete-webpack-plugin
+[core-js]:                                   https://github.com/zloirock/core-js
 
 ## Table of Contents
 


aunnop739:
#418

ai:
@zloirock thanks for sending PR. I can make a tweet about it.

Do you include polyfills depends on Browserslist config?

zloirock:
@ai here nothing new, I made it mainly for `@babel/preset-env` at the beginning of the year, so it works on `useBuiltIns`, `core-js-builder` and other third-party tools as a free bonus - sure, it's based on Browserslist config.

ai:
@zloirock are you allow to ignore your own `target` options for users and use global Browserslist config?

The whole idea of the Browserslist is to have a single config for target browsers, because:

1. It is easy to forget set target browsers for some tool (especially if you do not expect from tool to support different target browsers).
2. It is easy to forget to change target browsers everywhere in the big project.
3. It is hard for every separated tool to spread best practices for target browsers (and developers use very bad queries in reality). One config is a good thing to focus developers on thinking about what they are putting there.

zloirock:
@ai I know it. But, because it's (mainly) low-level tool, for don't break some use cases when it's ignored, provided all polyfills. The decision which targets should be used does a high-level tool like `preset-env`. However, since `babel` ruined our relationship, most likely I'll move high-level polyfilling tools to `core-js` side in the future.

ai:
@zloirock in this case, I need to remove Core JS from this list, since it works with Browserslist in a different way.

Feel free to ping me during the next major update to fix this problem.

zloirock:
@ai feel free to remove it. Just thought that you were interested in it -)

ai:
@zloirock yeap, it was interesting. Feel free to ask if you will any help or feature from Browserslist

PR #419, opened 2019-10-22
Look for cached config in each parent folder

Knagis:
Slight improvement for https://github.com/postcss/autoprefixer/issues/1256

Note that this also better matches the updating of cache entries, since eachParent does path.resolve, which the old code didn't do for reading cache


ai:
Do you have a twitter account to mention you in a tweet about the release?

Knagis:
No, I am not using twitter.

PR #425, opened 2019-11-28
change most prominent example to be more like listed best practice

dannyvankooten:
Because the README has a "Best Practices" section that states the following:

> If you want to change the default set of browsers we recommend to combine last 1 version, not dead with > 0.2% (or > 1% in US, > 1% in my stats). last n versions adds too many dead browsers and does not add popular old versions. Choosing a percentage above 0.2% will in the long run make popular browsers even more popular. We might run into a monopoly and stagnation situation, as we had with Internet Explorer 6. Please use this setting with caution.

I figured it make sense to update the most prominent example to be more like that best practice and promote browser diversity in the long run. (I think) A lot of people will just be copying that example in their package.json and think it is the default or recommend setting, so this should help combat that.

I've also moved the `default` setting to the top of the list just to show how absurd a browserlist setting of `> 5%` would be compared to the default.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,9 +19,9 @@ when you add the following to `package.json`:
 
 ```js
   "browserslist": [
-    "last 1 version",
-    "> 0.25%",
-    "maintained node versions",
+    "last 2 versions",
+    "> 0.5%",
+    "Firefox ESR",
     "not dead"
   ]
 ```
@@ -31,9 +31,9 @@ Or in `.browserslistrc` config:
 ```yaml
 # Browsers that we support
 
-last 1 version
-> 0.25%
-maintained node versions
+last 2 versions
+> 0.5%
+Firefox ESR
 not dead
 ```
 
@@ -104,10 +104,6 @@ Browserslist will take queries from tool option,
 
 ## Best Practices
 
-* Select browsers directly (`last 2 Chrome versions`) only if you are making
-  a web app for a kiosk with one browser. There are a lot of browsers
-  on the market. If you are making general web app you should respect
-  browsers diversity.
 * There is a `defaults` query, which gives a reasonable configuration
   for most users:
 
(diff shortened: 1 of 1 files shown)


ai:
I like the idea to make examples better!

> Choosing a percentage above 0.2% will in the long run make popular browsers even more popular

Can’t remember why I put `0.2%` there. I am not sure that there is a limit.

We use `last 2 versions, not dead` to support browsers' diversity. It will add even browsers with < 0.1%. Should we add this shortcut to examples instead?

dannyvankooten:
Whoa, you're quick @ai. And sure, that makes a lot of sense, I'll update the PR!

README.md
```diff
@@ -19,9 +19,9 @@ when you add the following to `package.json`:
 
 ```js
   "browserslist": [
-    "last 1 version",
-    "> 1%",
-    "maintained node versions",
```
ai: sorry, we need to promote `node` example, since nobody know about it

dannyvankooten:
I tried rewriting the "Best Practices" section a bit, but it striked me that removing the usage number from the example would mean excluding popular older versions (as stated in the original text), so while not directly affecting browser diversity, it is probably not desirable either, right?

So instead I changed the main example to be exactly like the default and rewrote the best practices section only to explain why to combine different configurations.

README.md
```diff
@@ -19,9 +19,9 @@ when you add the following to `package.json`:
...
+    "last 2 versions",
```
ai: This query is very close  to `defaults`. Maybe we can promote `defaults` here? For instance:

```
defaults
not IE 11
maintained node versions
```

ai:
At least for me “popular older versions” is not a big deal. I think old browsers are bad for industry and users (since they have a lack of security updates). In terms of diversity, we can force just the latest versions for each browser.

Of course, there is no rational way to select the best recommendations for Browserslist. I can change my mind if you will show that popular older browsers should be supported.

dannyvankooten:
I think that makes sense @ai. It showcases that you can inherit from `defaults` (which should be a sensible thing to do for most people), is slightly [more inclusive](https://browserl.ist/?q=defaults%2C+not+IE+11%2C+maintained+node+versions) than the [previous example](https://browserl.ist/?q=last+1+version%2C+%3E+1%25%2C+maintained+node+versions%2C+not+dead) and also showcases some lesser known but useful settings. I like it.

PR #426, opened 2019-12-01
chore: exclude IE_Mob 11 in readme examples

JLHwung:
The readme example demonstrates how to exclude `IE 11` from the defaults query. However, in this example `IE Mob` (thus, IE Mobile is literally IE Mob) is not excluded. It would cause trouble for those devtools (namely Babel/preset-env) that maps `IE Mob` to `IE` and thus applies IE-related polyfills, which is very confused for the end users since `not IE 11` has been specified. (C.f. https://github.com/babel/babel/issues/10635#issuecomment-549400423)

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,6 +21,7 @@ when you add the following to `package.json`:
   "browserslist": [
     "defaults",
     "not IE 11",
+    "not IE_Mob 11",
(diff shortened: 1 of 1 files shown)


ai:
Great idea! 👍

PR #428, opened 2019-12-04
Fix android queries

JLHwung:
I recommend to review this PR by commits.

The first commits merge chrome >= 37 data to android under the `mobileToDesktop` flag

The second commits fixed result sorting issues occured when I wrote test for the first commit.

I propose in In the next major release we should set the defaults of `opts.mobileToDesktop` to be `true` so that the returned mobile browsers are complete. People can disable this feature if they find things are not going well.

### DIFF
--- a/index.js
+++ b/index.js
@@ -7,7 +7,7 @@ var e2c = require('electron-to-chromium/versions')
 var BrowserslistError = require('./error')
 var env = require('./node') // Will load browser.js in webpack
 
-var FLOAT_RANGE = /^\d+(\.\d+)?(-\d+(\.\d+)?)*$/
+var FLOAT_RANGE = /^\d+(\.\d+){0,2}?(-\d+(\.\d+){0,2}?)*$/
 var YEAR = 365.259641 * 24 * 60 * 60 * 1000
 
 var QUERY_OR = 1
@@ -138,9 +138,9 @@ function compare (a, b) {
 
 function compareSemver (a, b) {
   return (
-    compare(a[0], b[0]) ||
-    compare(a[1], b[1]) ||
-    compare(a[2], b[2])
+    compare(parseInt(a[0]), parseInt(b[0])) ||
+    compare(parseInt(a[1] || '0'), parseInt(b[1] || '0')) ||
+    compare(parseInt(a[2] || '0'), parseInt(b[2] || '0'))
   )
 }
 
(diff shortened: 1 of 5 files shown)


index.js
```diff
@@ -367,11 +399,13 @@ function browserslist (queries, opts) {
...
     name2 = name2.split(' ')
     if (name1[0] === name2[0]) {
-      if (FLOAT_RANGE.test(name1[1]) && FLOAT_RANGE.test(name2[1])) {
-        return parseFloat(name2[1]) - parseFloat(name1[1])
-      } else {
-        return compare(name2[1], name1[1])
-      }
+      // here we assume that caniuse version ranges never overlaps,
+      // so it is safe to use the left of the range
+      // eslint-disable-next-line max-len
```
ai: Nope, you can't just disable the rule. Try to reorganize the code to make it more readable.

test/range.test.js
```diff
@@ -8,6 +8,12 @@ beforeEach(() => {
       name: 'ie',
       released: ['8', '9', '10', '11'],
       versions: ['8', '9', '10', '11']
+    },
+    android: {
+      name: 'android',
+      released: ['4.2-4.3', '4.4',
```
ai: Strange new line

test/range.test.js
```diff
@@ -24,6 +30,13 @@ it('selects versions with query out of range', () => {
   expect(browserslist('ie 1-12')).toEqual(['ie 11', 'ie 10', 'ie 9', 'ie 8'])
 })
 
+it('selects a range of android browsers', () => {
+  expect(browserslist('android 4.3-37'))
+    .toEqual([
```
ai: You can keep `toEqual` on previous line to make method smaller

index.js
```diff
@@ -191,14 +191,42 @@ function byName (name, context) {
...
+      var cloned = cloneData(desktopData)
+      cloned.name = name
+      return cloned
+    }
   }
   return browserslist.data[name]
 }
 
+function normalizeAndroidVersions (androidVersions, chromeVersions) {
+  var firstEvergreen = 37
```
ai: If we use this bar in multiple places, we need to move it to `var ANDROID_FIRST_EVERGREEN` constant.

test/range.test.js
JLHwung: 😛I am really not good at formatting without the help of prettier. If you are okay with prettier, I would suggest we leave it as-is. I can introduce prettier in a separate PR.

ai: Don't worry, fix the logic and `// eslint-disable` and I will fix the formatting.

test/range.test.js
JLHwung: Good catch, addressed in bd31b51.

ai:
Thanks. Released in 4.8.1.

PR #448, opened 2020-02-13
fix last N major versions with non-sequential version numbers

### LINKED ISSUE #447: "last N major versions" ignoring breaks in non-sequential version numbers
Right now something like "last 4 edge major versions" only reports 79 and 80, ignores 17 and 18. "last 4 edge versions" reports all four correctly.

**Using version**: 4.8.6

**Steps to reproduce**:
```
const browsers = require('browserslist');
browsers('last 4 edge major versions'); // [ 'edge 80', 'edge 79' ]
browsers('last 4 edge versions'); // [ 'edge 80', 'edge 79', 'edge 18', 'edge 17' ]
browsers('last 64 edge major versions'); // [ 'edge 80', 'edge 79', 'edge 18', 'edge 17' ]
```

**Expected behavior**: report 17, 18, 79, and 80 as last 4 major versions of edge.
**Actual behavior**: only 79 and 80 are reported

vetruvet:
Fixes #447.

This changes the way the minimum major version is calculated.
Previously it was calculated by subtracting the desired number of versions from the maximum version number which does not yield the desired outcome when a browser has non-sequential version numbers (e.g. edge 17, 18, 79, 80).

The change involves finding the released major versions and taking the last N of those rather than relying on simple subtraction and an assumption of sequential version numbers.

### DIFF
--- a/index.js
+++ b/index.js
@@ -53,7 +53,11 @@ function getMajor (version) {
 
 function getMajorVersions (released, number) {
   if (released.length === 0) return []
-  var minimum = getMajor(released[released.length - 1]) - parseInt(number) + 1
+  var majorVersions = uniq(released.map(getMajor))
+  var minimum = majorVersions[majorVersions.length - number]
+  if (!minimum) {
+    return released
+  }
   var selected = []
   for (var i = released.length - 1; i >= 0; i--) {
     if (minimum > getMajor(released[i])) break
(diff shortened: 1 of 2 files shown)


ai:
Thanks. Looks good.

I am going to sleep right now and will accept and release it today.

ai:
Thanks. Released at 4.8.7.

PR #450, opened 2020-02-14
Add browserslist --update-db for caniuse-lite updating

Ibadichan:
https://cultofmartians.com/tasks/browserslist-update-db.html


ai:
If you will have to improve `--help`, it will be nice.

But you can do it after finishing the main part and marked as the task solver.

ai:
@Ibadichan You should print origin errors (in file operation and CLI calls) since without the original message will not be able to understand the reason

Ibadichan:
> @Ibadichan You should print origin errors (in file operation and CLI calls) since without the original message will not be able to understand the reason

Do you want to say that i must use BrowserslistError function to generate errors? What do you mean by saying  "origin errors"?

ai:
```js
fs.readFile(file, (err, content) => {
  if (err) throw err
```

Will print the origin error to the console. This message is more useful, than `an reading file error occurred`, since the last one do not explain what happened and how to fix the problem.

test/updateDB.test.js
```diff
@@ -0,0 +1,197 @@
+const path = require('path')
```
ai: We use dashed file names like [`update-db.test.js`](https://github.com/browserslist/browserslist/blob/master/test/shareable-stats.test.js)

test/updateDB.test.js
```diff
@@ -0,0 +1,197 @@
...
+const execSync = require('child_process').execSync
+
+const updateDB = require('../updateDB')
+const BrowserslistError = require('../error')
+
+const root = process.cwd()
+const lastVersion = getLastVersion()
+
+beforeEach(() => {
+  jest.spyOn(console, 'log').mockImplementation(() => true)
```
ai: It will be better to mock `console.log` only in a specific test, since I may use `console.log` to debug some tests in the future.

test/updateDB.test.js
```diff
@@ -0,0 +1,197 @@
...
+})
+
+afterEach(() => {
+  process.chdir(root)
+})
+
+describe('updateDB function', () => {
```
ai: We do not use `describe` in other tests

test/updateDB.test.js
```diff
@@ -0,0 +1,197 @@
...
+
+  it('prints the latest version of caniuse-lite', async () => {
+    let npmPackage = await createNpmPackage()
+    let message = 'Fetching caniuse-lite ' + lastVersion + '...'
+
+    process.chdir(npmPackage)
+
+    updateDB()
+
+    expect(console.log).toHaveBeenCalledWith(message)
```
ai: It will be more readable if you will put a message to the `expect`, rather than using variable

test/updateDB.test.js
```diff
@@ -0,0 +1,197 @@
...
+      'utf8'
+    )
+
+    let content = newYarnLock.split('\n').join('')
+
+    // eslint-disable-next-line
+    let regExp = new RegExp(`caniuse-lite@.*  version "${ lastVersion }"`)
+
+    expect(yarnLock).not.toEqual(newYarnLock)
+    expect(regExp.test(content)).toEqual(true)
```
ai: It is better to use `expect(content).toMatch(/regexp/)`

updateDB.js
```diff
@@ -0,0 +1,88 @@
...
+
+  function startsWithTwoSpaces (line) {
+    return /^ {2}/.test(line)
+  }
+
+  var lockfile, packageManager
+
+  var packageJson = findUp.sync('package.json')
+  if (!packageJson) {
+    throw new BrowserslistError('Cannot find package.json')
```
ai: Bad error message. You do not explain what happened and how to fix it.

For instance, user can call Browserslist from wrong dir.

updateDB.js
```diff
@@ -0,0 +1,88 @@
...
+  }
+
+  var npmLockfile = path.join(packageJson, '/../package-lock.json')
```
ai: Use `join(dirname(packageJson), 'package-lock.json')`. `..` is not clear here.

updateDB.js
```diff
@@ -0,0 +1,88 @@
...
+
+  if (fs.existsSync(npmLockfile)) {
+    lockfile = npmLockfile
+    packageManager = 'npm'
+  } else if (fs.existsSync(yarnLockfile)) {
+    lockfile = yarnLockfile
+    packageManager = 'yarn'
+  } else {
+    throw new BrowserslistError(
+      'No lockfile found. Run "yarn install" or "npm install"'
```
ai: This is a good error message

updateDB.js
```diff
@@ -0,0 +1,88 @@
...
+    )
+  }
+
+  console.log('Fetching caniuse-lite ' + getLastVersion() + '...')
```
ai: Use `…` special symbol for ellipsis

updateDB.js
```diff
@@ -0,0 +1,88 @@
...
+
+  var parsedContent, newContent
+  var content = fs.readFileSync(lockfile, 'utf8')
+
+  if (packageManager === 'npm') {
+    parsedContent = JSON.parse(content)
+
+    if (parsedContent && parsedContent.dependencies) {
+      delete parsedContent.dependencies['caniuse-lite']
```
ai: Do you update all nested `caniuse-lite` versions? It could be many different `caniuse-lite` on different nested levels (if I am right about nested nature of `package-lock.json`)

badimalex: Thank you for code review, i will update this function.

Ibadichan: Thank you for code review, i will update this function.

ai:
This PR was first and finished earlier https://github.com/browserslist/browserslist/pull/455/

PR #454, opened 2020-02-18
feat: add support of range syntax for node

### LINKED ISSUE #452: `node 8-12` throws `browserslist: Unknown browser node`
Not sure if related to #410. When querying `node 8-12` (or anything in a similar fashion, e.g. `node 10-13`), it says `browserslist: Unknown browser node`.

### Expected behavior
I would say `node 8-12` should return all minor versions from 8.0.0 to the latest version in major version 12 (currently 12.16.0).

However, it conflicts with the existing behavior of `node 8` selecting only the latest minor version (8.17.0 but not 8.0.0), which may cause confusion.

### Workaround
In the first case (`node 8-12` returns all versions starting 8.0.0), you can use `node >=8 and node < 13` instead.

zypA13510:
Close #452.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
 The MIT License (MIT)
 
-Copyright 2014 Andrey Sitnik <andrey@sitnik.ru>
+Copyright 2014 Andrey Sitnik <andrey@sitnik.ru> and other contributors
 
 Permission is hereby granted, free of charge, to any person obtaining a copy of
 this software and associated documentation files (the "Software"), to deal in
--- a/index.js
+++ b/index.js
@@ -148,6 +148,43 @@ function compareSemver (a, b) {
   )
 }
 
+// this follows the npm-like semver behavior
+function semverFilterLoose (operator, range) {
+  range = range.split('.').map(parseSimpleInt)
+  if (typeof range[1] === 'undefined') {
+    range[1] = 'x'
+  }
+  // ignore any patch version because we only return minor versions
+  // range[2] = 'x'
+  switch (operator) {
+    case '<=':
+      return function (version) {
+        version = version.split('.').map(parseSimpleInt)
+        return compareSemverLoose(version, range) <= 0
+      }
+    default:
+    case '>=':
+      return function (version) {
+        version = version.split('.').map(parseSimpleInt)
+        return compareSemverLoose(version, range) >= 0
+      }
+  }
+}
+
+// this follows the npm-like semver behavior
+function compareSemverLoose (version, range) {
+  if (version[0] !== range[0]) {
+    return version[0] < range[0] ? -1 : +1
+  }
+  if (range[1] === 'x') {
+    return 0
+  }
+  if (version[1] !== range[1]) {
+    return version[1] < range[1] ? -1 : +1
+  }
+  return 0
+}
+
 function resolveVersion (data, version) {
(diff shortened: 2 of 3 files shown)


zypA13510:
I think the size limit is exceeded due to the added dependency on 'semver'. Do you want to relax the size limit? Or do I need to reimplement semver?

ai:
I can’t add only one contributor to `package.json`. We can try to add all contributors, but in this case, we need a scalable automatic solution and it will increase the npm package size a lot.

ai:
> I think the size limit is exceeded due to the added dependency on 'semver'. Do you want to relax the size limit?

Nope, we need to keep Browserslist small since webpage size is a big problem nowadays

zypA13510:
> I can’t add only one contributor to `package.json`. We can try to add all contributors, but in this case, we need a scalable automatic solution and it will increase the npm package size a lot.

I don't think an automatic solution is possible, or accurate (which name to use, email, URL, etc.). Maybe it's easier to just skip this contributors part. I'll remove it.

---

As for the size problem,

> since webpage size is a big problem nowadays

is this relevant? From my experience, and the "It is used in" list on README, most of the tools that use browserslist run on the server-side, during build/transpilation, and it's not served directly to the client/browser. In another word, for a user-facing web project, this package is most likely somewhere in the "devDependencies", rather than "dependencies", is it not?

ai:
You are totally right, that there are no good popular user cases to use Browserslist on client-side. But there are 2 problems:

1. A lot of users still use it on the client-side. They use it without good reasons. But if we will increase the size, clients will suffer.
2. There are good reasons to use Browserslist on the client-side. For instance, interactive tools like [Online Autoprefixer](https://autoprefixer.github.io/).

Do we have any other libraries with the same features, but a smaller size?

zypA13510:
Ok, I removed the dependency on semver and it should pass the tests now.

ai:
Thanks. Released at 4.9

zypA13510:
@ai btw, back to the expected behavior question. I still feel that it's strange for `node 6-8` to include `node 8.x.0`, but `node <= 8` to max out at `node 8.0.0`. And I've also identified how can I fix this (make `node <= 8` to also contain `node 8.x.0`). However, I'm afraid this would be a breaking change, are you interested in "fixing" this?

ai:
I think it is expected that `node 6-8` should contain `node 6` (the latest node 6.x) and `node 8` results.

`node < 8` should not include `node 8.0.0` and `<=` should add only one extra version.

I think the current behavior is OK. Changing `<=` behavior will create even more questions.

zypA13510:
Why I'd consider `node <= 8` to include `8.1.0`:

Because NPM does this. The semver implementation used by NPM is [here](https://www.npmjs.com/package/semver), and in this implementation, `<= 8` actually includes every `8.x.x` version. And the place we do browserslist configuration, `package.json`, is actually a file managed by NPM.

Maybe I should create a new issue and let more people weigh in on this? What do you think?

zypA13510:
Actually, on a second thought I think including only `8.0.0` may be fine. Because usually `8.1.0` will not introduce any breaking change, and what's optimized/polyfilled for `8.0.0` will work just fine.

ai:
Another problem, that `safari <= 10` contains only `10.0`, not `10.1`.

Let’s keep the origin behavior. It is a very rare query. I do not know good use case for `node <= X`

PR #455, opened 2020-02-23
Add the --update-db flag

isolovev:
Hi. PNPM was not mentioned in the task, but it is being developed and used in some projects. I also added an update for it.


ai:
1. PNPM support 👍
2. `--update-db` is a rare feature. Let’s not add dependencies for this case.
3. `browserslist --update-db` could be run in project subdir (for instance, `test/`)
4. Let’s throw origin error. A message like `A reading Lockfile error occurred` does not describe what happened.
5. I do not use structures like `PM`. Let’s use `pm = 'yarn'` to keep consistency with another code.
6. Should we rename `cleanupLockfile` to `updateLockfile`?
7. `data` var is not specific.
8. In the future, we have plan for `caniuse-lite-json` npm package. We can’t use `^caniuse-lite` check.

isolovev:
Can I possibly use such a function so as not to break the consistency of the code?
```
function isYarn () {
  return packageManager === 'yarn'
}
```

ai:
Why not `if (packageManager === "yarn")` without additional function?

isolovev:
The `pkg-up` package was specified in the task. Can I use it?

ai:
Yeap, let’s use this dependency. It is small, used in other packages and really helpful for us.

updater.js
```diff
@@ -0,0 +1,252 @@
...
+var fs = require('fs')
+var childProcess = require('child_process')
+var pkgUp = require('pkg-up')
+
+var BrowserslistError = require('./error')
+
+var UNKNOWN_VERSION = 'unknown version'
+var PACKAGE_CANIUSE = 'caniuse-lite'
+
+var packageManager = ''
```
ai: Using a global state is not good for simple functions like we have in this task.

The code will be more readable if we will pass it to functions as an argument.

updater.js
```diff
@@ -0,0 +1,252 @@
...
+      console.log(
+        'You have the latest version of ' + PACKAGE_CANIUSE + ' installed'
+      )
+      return
+    }
+
+    console.log(
+      'Current version: ' + currentVersion + '\n' +
+      'New version: ' + lastVersion + '\n' +
+      'Updating ' + PACKAGE_CANIUSE + '...'
```
ai: 3 dots is not equal to `…` special semantic symbol.

updater.js
```diff
@@ -0,0 +1,252 @@
+var path = require('path')
+var fs = require('fs')
+var childProcess = require('child_process')
+var pkgUp = require('pkg-up')
+
+var BrowserslistError = require('./error')
+
+var UNKNOWN_VERSION = 'unknown version'
```
ai: Why do we need these constants?

updater.js
```diff
@@ -0,0 +1,252 @@
...
+
+  if (fs.existsSync(lockfilePnpm)) {
+    packageManager = 'pnpm'
+    return lockfilePnpm
+  }
+
+  throw new BrowserslistError('Lockfile search error')
+}
+
+function getPackageManager () {
```
ai: Why do you need this function and can’t use `packageManager` directly?

updater.js
```diff
@@ -0,0 +1,252 @@
...
+    packageManager = 'yarn'
+    return lockfileYarn
+  }
+
+  if (fs.existsSync(lockfilePnpm)) {
+    packageManager = 'pnpm'
+    return lockfilePnpm
+  }
+
+  throw new BrowserslistError('Lockfile search error')
```
ai: A good error message explains why it happened and how to fix iy

updater.js
```diff
@@ -0,0 +1,252 @@
...
+      'Error getting the latest version of ' + PACKAGE_CANIUSE
+    )
+  }
+
+  return versions[0]
+}
+
+function updateLockfile (lockfileRaw) {
+  if (packageManager === 'yarn') {
+    return cleanupYarnLock(lockfileRaw)
```
ai: If you use `updateLockfile` name, you should use `updateYarnLock`, etc

ai: I think it will be better to keep all code in one function in `if (packageManager === 'yarn') {` without functions for each package managers

updater.js
```diff
@@ -0,0 +1,252 @@
...
+
+var UNKNOWN_VERSION = 'unknown version'
+var PACKAGE_CANIUSE = 'caniuse-lite'
+
+var packageManager = ''
+
+function updateDB () {
+  var lockfile = getLockfile()
+
+  fs.readFile(lockfile, 'utf8', function (readError, lockfileRaw) {
```
ai: You use sync methods in one case and async methods in another. Let’s be consistent here.

I think sync methods are OK here since there will be no other task run in the same Node.js instance (async tasks are good when you have many different tasks, like in Gulpfile or web server)

updater.js
```diff
@@ -0,0 +1,252 @@
...
+    if (readError) {
+      throw readError
+    }
+
+    var currentVersion = getCurrentVersion(lockfileRaw)
```
ai: It could be many different versions of `caniuse-lite` inside lockfile (in subdependencies). Are you sure that you are getting the minimum version here and updating all versions (including merging different versions)?

updater.js
```diff
@@ -0,0 +1,252 @@
...
+        }
+
+        console.log(PACKAGE_CANIUSE + ' has been successfully updated')
+      })
+    })
+  })
+}
+
+function getLockfile () {
+  var packagePath = pkgUp.sync()
```
ai: You forget the case then `package.json` was not found

isolovev: I agree. This case needs to be covered.

isolovev:
I agree that need to expand the description for the user.

isolovev:
Do you think? Do I need to separate the lockfiles update logic into different files? That behavior changes occur in isolation. It will also be easier to expand it.

ai:
> Do you think? Do I need to separate the lockfiles update logic into different files? 

Nope, I think we should keep everything in one file.

isolovev:
I'm still doing the task.

ai:
@isolovev can you remove `merge` commit? It make diff unreadable.

You can remove commit with `git reset HEAD^ && git checkout . && git push …`

isolovev:
Ok. I’ll remove to morning.

isolovev:
@ai repeat review pls. I fixed your comments.
It is also possible to optimize the code of the `updateLockfile` function. Use a single loop and perform checks in it. Some code will be shared.

ai:
Tomorrow. Today is a day for crepes party =^_^=

cli.js
```diff
@@ -3,6 +3,7 @@
 var fs = require('fs')
 
 var browserslist = require('./')
+var updateDB = require('./updater')
```
ai: The function should have the same name as a file.

I suggeest `updateDB = require('./update-db')` since we import function and it should be verb.

test/release.test.js
```diff
@@ -138,14 +134,3 @@ it('shows warning only once', () => {
   browserslist('last 2 versions')
   expect(console.warn).toHaveBeenCalledTimes(1)
 })
-
```
ai: I think it will be very good to have tests for `--update-db`. It will help us to be sure about what we are doing.

updater.js
```diff
@@ -0,0 +1,225 @@
+var path = require('path')
+var fs = require('fs')
+var childProcess = require('child_process')
+var pkgUp = require('pkg-up')
+
+var BrowserslistError = require('./error')
+
+var PACKAGE_CANIUSE = 'caniuse-lite'
```
ai: Let’s remove this constant. I do not know a reason to have have `caniuse-lite` directly in strings and code

updater.js
```diff
@@ -0,0 +1,225 @@
...
+  if (packageManager === 'npm') {
+    var parsedFile = JSON.parse(lockfileRaw)
+    if (
+      parsedFile.dependencies !== undefined &&
+      parsedFile.dependencies[PACKAGE_CANIUSE] !== undefined &&
+      parsedFile.dependencies[PACKAGE_CANIUSE].version !== undefined
+    ) {
+      return parsedFile.dependencies[PACKAGE_CANIUSE].version
+    }
+    return null
```
ai: `Current version is null` will be bad output.

I think you need better message for this case.

updater.js
```diff
@@ -0,0 +1,225 @@
...
+  }
+
+  var parsedLockfileLines = lockfileRaw.split('\n')
+  var countLines = parsedLockfileLines.length
+  if (packageManager === 'yarn') {
+    var foundCaniuse = false
+    for (var i = 0; i < countLines; i++) {
+      var row = parsedLockfileLines[i]
+      if (row.indexOf(PACKAGE_CANIUSE) === 0) {
+        foundCaniuse = true
```
ai: Code with variable and `continue` is too tricky :(. Are you need the next row? This code can be easy:

```js
      if (row.indexOf('caniuse-lite') === 0) {
        vat nextRow = parsedLockfileLines[i + 1]
        var rowVersion = /version "([^"]+)"/.exec(nextRow =)
        if (rowVersion.length > 1) {
          return rowVersion[1]
        }
```

updater.js
```diff
@@ -0,0 +1,225 @@
...
+  }
+}
+
+function updateLockfile (lockfileRaw, info, packageManager) {
+  var parsedLockfileLines = lockfileRaw.split('\n')
+  var countLines = parsedLockfileLines.length
+  var hasNecessaryDeps = false
+
+  if (packageManager === 'npm') {
+    for (var i = 0; i < parsedLockfileLines.length; i++) {
```
ai: Should be work with `package-lock.json` as with JSON? It will be simpler and safer code.

updater.js
```diff
@@ -0,0 +1,225 @@
...
+        parsedLockfileLines[j].indexOf('@') > 0
+      ) {
+        hasNecessaryDeps = true
+      }
+
+      if (!hasNecessaryDeps) {
+        continue
+      }
+
+      if (parsedLockfileLines[j].indexOf('version') > 0) {
```
ai: I suggest you use `j++` to go to next line instead of goind to next step by `for` with some state-variable.

```
for (var i = 0; i < lines.legnth; i++) {
  if (test(line[i])) {
    line[i + 1] = fix2
    line[i + 2] = fix2
    i += 2
  }
}
```

updater.js
isolovev: If you plan to use a package with a different name, such as `caniuse-lite-json`, then I should leave it and not inline it is code.

updater.js
isolovev: I agree that the work will be safer, but the search for nested dependencies will be complicated by tree smash. I try that improvement.

updater.js
ai: Rare case. In this case I will replace it with Find-and-Replace in my text editor.

updater.js
ai: You can try recursion to make it easy:

```js
function process (node) {
  if (check(node) {
    node = changeNode(node)
  }
  node.children = node.children.map(i => process(i))
  return node
}
```

isolovev:
@ai need review pls.
I fixed comments and added tests.

package.json
```diff
@@ -24,6 +24,7 @@
     "caniuse-lite": "^1.0.30001030",
     "electron-to-chromium": "^1.3.363",
     "node-releases": "^1.1.50"
+    "pkg-up": "^3.1.0"
```
JLHwung: `pkg-up` requires `node >= 8` support. `browserslist` is depended by `@babel/preset-env` which still offers `node >= 6` support. Are there any chance that we may downgrade `pkg-up` to `^2.0.0`?

ai: Sure ae5e3a6

ai: Released at 4.11.1

PR #458, opened 2020-02-29
Workaround incorrect caniuse regional data

### LINKED ISSUE #431: `mobileToDesktop` breaking change between 4.7.3 and 4.8.2?
`require('browserslist')("cover 90% in CN", { mobileToDesktop: true })` gives `and_chr 78` and other 4 items in 4.7.3, but gives `and_chr 4` and other 4 items in 4.8.2.

We're using `@babel/preset-env` and it broke recently after update. I traced and found `browserslist` gives different result before and after. So I come with the line above. `@babel/preset-env` always calls `browserslist` with `mobileToDesktop: true`. 

Honestly I cannot quite understand how `mobileToDesktop` works from `README.md`, but I found several changes regarding `mobileToDesktop` option between 4.7.3 and 4.8.2. Is current result(4.8.2) correct?

JLHwung:
In this PR we normalize the usage data
```js
{ and_chr: {0: 30.6943} }
```
returned from the `caniuse-db` to
```js
{ and_chr: {80: 30.6943} }
```
where `80` is the latest version in `browserslists.data`. By doing so the regional data should not return `chrome 4` for the invalid `and_chr 0` query.

Fixes #431

### DIFF
--- a/index.js
+++ b/index.js
@@ -588,7 +588,7 @@ browserslist.coverage = function (browsers, stats) {
     } else {
       stats = stats.toUpperCase()
     }
-    env.loadCountry(browserslist.usage, stats)
+    env.loadCountry(browserslist.usage, stats, browserslist.data)
     data = browserslist.usage[stats]
   } else {
     if ('dataByBrowser' in stats) {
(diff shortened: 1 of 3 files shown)


ai:
You are doing really great PRs :+1:

PR #462, opened 2020-03-22
feat: add mobileToDesktop option to CLI

Cherry:
This PR adds the `--mobileToDesktop` or `-m` option to the `cli.js`. 

This was primarily added to test this functionality without using the JS API. A real-world use-case I had recently was wanting to easily and quickly test browserslist output as per the issue at https://github.com/browserslist/browserslist/issues/431

I'm happy to make any changes you feel are necessary to get this merged.

### DIFF
--- a/cli.js
+++ b/cli.js
@@ -18,6 +18,7 @@ var USAGE = 'Usage:\n' +
             '  npx browserslist --coverage=US,RU,global "QUERIES"\n' +
             '  npx browserslist --env="environment name defined in config"\n' +
             '  npx browserslist --stats="path/to/browserlist/stats/file"\n' +
+            '  npx browserslist --mobileToDesktop\n' +
             '  npx browserslist --update-db'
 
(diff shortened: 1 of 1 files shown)


ai:
Thanks for PR!

Bad news: I removed `-m` alias. It is a rare use case and I prefer to keep short aliases free for future big features. c9d5965

Neutral news: I renamed argument to `--mobile-to-desktop` to clean name convention (like `--update-db`) f3ac335

Good news: your fix was already released at 4.11. It was a good idea, this argument can be useful. I share your idea, that we should think about the non-JS world during Browserslist development.

Cherry:
Thanks! I wasn't sure about the exact convention to follow there, but I'm glad this was merged. 😀

I think I'm still tracking down a similar bug to #431 here, but this time with `browserslist-stats` data and `mobileToDesktop` normalization, but I'll keep investigating that and open another issue once I have more solid info.

Issue #300, opened 2018-09-18
Support environments for shareable configs


Fer0x:
I would like to keep my shareable config similarly to package.json field "browserslist", like this:

```js
module.exports = {
  development: ['last 2 chrome versions', 'last 2 firefox versions', 'last 2 edge versions'],
  production: [
    '>1%',
    'last 3 versions',
    'not dead',
    'not op_mini all',
    'Firefox ESR',
    'not ie < 11',
  ],
};
```

For now it's impossible because of error: `config exports not an array of queries `.

ai:
Good idea.

I am not sure when I will fix it  (because most of tools doesn’t support environments, it is very advanced and rare feature). If you want it quicker, you can send PR.

Fer0x:
@ai ok, I'll take a look at it soon.

JamiesonRoberts:
@Fer0x Could the same pattern as eslint, prettier, and babel be used, where you can have use a `.browserlistrc.js` or `browserlist.config.js` file in your repo? This would keep it fitting the file patterns despite being different projects.

I know for example Prettier and I think esLint use Comsiconfig (https://github.com/davidtheclark/cosmiconfig) to handle finding the files. Happy to take a stab at it if you haven't done it yet.

ai:
@JamiesonRoberts Browserslist can’t use JS configs and Comsiconfig, because Browserslist config should be able to be used by non-JS tools (like AutoprefixerRails has Ruby code to load Browserslist config)

JamiesonRoberts:
@ai Maybe fair about cosmiconfig, but it should be possible to handle the array from a module.export the same as the package.json version. Ultimately as long as it exports an array would it not be possible to load from a .js file?

ai:
@JamiesonRoberts if you want just `dev` vs `production` you can use Browserslist environments:
https://github.com/browserslist/browserslist#environments

Maybe it will be enough. Or you want completely dynamically queries?

JamiesonRoberts:
@ai my preference is dynamic queries or setting things programmatically rather than relying on specific key in an object to match. That way I can retrieve the values in several different scenarios, rather than just simple ones like above.

ai:
> my preference is dynamic queries or setting things programmatically

1. I do not like to add features without real use case. Often you just add features and nobody will use them later (but you still need to support them).
2. How Autoprefixer Rails (written on Ruby) will read this config?

JamiesonRoberts:
[...]

PR #474, opened 2020-04-19
✨ shareable: env support

yepninja:
Resolves: #300

Example:
```js
module.exports = {
    defaults: ['> 1%],
    test: ['current node']
}
``` 

It's not breaking changes according to code.
But it's breaking changes according to tests, which make sense that I have to edit test case for non-array error.

### DIFF
--- a/node.js
+++ b/node.js
@@ -156,12 +156,17 @@ module.exports = {
     if (!context.dangerousExtend) checkExtend(name)
     // eslint-disable-next-line security/detect-non-literal-require
     var queries = require(require.resolve(name, { paths: ['.'] }))
-    if (queries && queries.constructor === Object) {
-      if (!queries.defaults) queries.defaults = []
-      return pickEnv(queries, context)
+    if (!queries) {
+      throw new BrowserslistError(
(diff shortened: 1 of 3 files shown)


node.js
```diff
@@ -156,12 +156,17 @@ module.exports = {
     if (!context.dangerousExtend) checkExtend(name)
     // eslint-disable-next-line security/detect-non-literal-require
     var queries = require(require.resolve(name, { paths: ['.'] }))
-    if (!Array.isArray(queries)) {
-      throw new BrowserslistError(
-        '`' + name + '` config exports not an array of queries'
-      )
+    if (queries && queries.constructor === Object) {
```
ai, line 159: What do we check here?

I do not know this pattern (so other developers will not understand it too)

ai: What if we will have

```js
if (Array.isArray(value)) {
} else if (typeof value === "object") {
}
```

yepninja: It would be enough if we will check before that value is not empty

ai:
I will wait for other docs PR and will release it in 4.12

yepninja:
Cool! Thank you for quick review. 🎉

Issue #435, opened 2020-01-09
[Feature] `safe`, `unsafe`, `strict`


MisterTicot:
The current defaults includes `Chrome 49` and `Safari 5.1` which as far as I can tell haven't been updated since more than two years. Shouldn't those be filtered by `not dead`?

> Safari for Windows
Apple no longer offers Safari updates for Windows. Safari 5.1.7 for Windows was the last version made for Windows, and it is now outdated.
>
> Published Date: September 24, 2018

https://support.apple.com/en-us/HT204416

ai:
`dead` was created to remove non-popular old browsers from `last 2 versions`.

Since this browsers were added by market share `> 1%` query, we can't just remove them (without changing of `dead` definition).

I think we should just wait until market share will go down.

MisterTicot:
Thanks for the explanation.

I understand that XP users don't have any other choices, and the point of browser.list is to have them included. I have two issues with this situation:

* The cost of supporting those outdated browsers is high, especially in term of polyfills.
* Those are unsafe to use.

I'm writing cryptocurrency apps - and I'd rather avoid having them run into unsafe browsers. I don't think I'm the only one who needs to mind about this. Maybe a `not unsafe` instruction could help, or an `unsafe` one that would have to be used to allow those browsers ?

ai:
Completely agree that it is not secure. But changing `dead` definition will require major release and for Browserslist it is really painful.

Can we think about cryptosecurity app as an non-standart case when you can change default browsers?

MisterTicot:
> Can we think about cryptosecurity app as an non-standart case when you can change default browsers?

Absolutely, but more generally my point is valid for any service connected to payments or secure applications. I agree that changing `dead` would be both confusing & annoying. For now, introducing an `unsafe` keyword that'll let us exclude out-of-date browsers seems to be the path of least resistance.

I'm not sure how difficult it would be? I assume that if it covers `default` browsers only, maybe as a first step, it would already be of great use.  `default, not unsafe`. Boom.

ai:
> For now, introducing an unsafe keyword that'll let us exclude browsers that don't receive security updates seems to be the path of least resistance.

I like the idea to add a new query.

The best way to do it is to have a list of pages for each browser (or at least main browsers) about what versions have security updates. I will manually check pages and update the query. Can you help with me here until I am on vacation?

MisterTicot:
[...]

PR #475, opened 2020-04-19
Add keywords: safe, unsafe, balanced

MisterTicot:
This PR implements the proposal that solves #435 by implementing new keywords: `safe`, `unsafe` as well as the browsers set `balanced` that greatly reduce poly-filling but excluding unsafe browsers while retaining a global coverage > 85%. ([analysis](https://github.com/browserslist/browserslist/issues/435#issuecomment-575407090))

### DIFF
--- a/README.md
+++ b/README.md
@@ -126,6 +126,11 @@ Browserslist will take queries from tool option,
     ]
   ```
 
+* If you're targeting modern browsers, we recommend using `balanced`. This set
+  cuts the bloat by selecting only browsers that still receive security updates.
+  It requires 75% less poly-filling than the `default` set while providing a
+  good global coverage (> 85%).
+  [(analysis)](https://github.com/browserslist/browserslist/issues/435#issuecomment-575407090)
 * If you want to change the default set of browsers, we recommend combining
   `last 2 versions`, `not dead` with a usage number like `> 0.2%`. This is
   because `last n versions` on its own does not add popular old versions, while
@@ -155,7 +160,7 @@ from one of these sources:
 4. `BROWSERSLIST` environment variable.
 5. If the above methods did not produce a valid result
    Browserslist will use defaults:
-   `> 0.5%, last 2 versions, Firefox ESR, not dead`.
+   `> 0.5%, last 2 versions, safe, not dead`.
 
 
 ### Query Composition
@@ -187,7 +192,9 @@ in your terminal._
 You can specify the browser and Node.js versions by queries (case insensitive):
 
 * `defaults`: Browserslist’s default browsers
-  (`> 0.5%, last 2 versions, Firefox ESR, not dead`).
+  (`> 0.5%, last 2 versions, safe, not dead`).
+* `balanced`: target modern (updated) browsers (`last 2 versions, safe, not
+  unsafe, not < 0.1%`).
 * `> 5%`: browsers versions selected by global usage statistics.
   `>=`, `<` and `<=` work too.
 * `> 5% in US`: uses USA usage statistics. It accepts [two-letter country code].
@@ -222,6 +229,10 @@ You can specify the browser and Node.js versions by queries (case insensitive):
 * `dead`: browsers without official support or updates for 24 months.
   Right now it is `IE 10`, `IE_Mob 10`, `BlackBerry 10`, `BlackBerry 7`,
   `Samsung 4` and `OperaMobile 12.1`.
+* `safe`: browsers that are not part of `last 2 versions` anymore but that are
+  still receiving security patches.
+* `unsafe`: browsers that are still part of `last 2 versions` but that are not
+  receiving security patches anymore.
 * `last 2 versions`: the last 2 versions for *each* browser.
 * `last 2 Chrome versions`: the last 2 versions of Chrome browser.
 * `not ie <= 8`: exclude browsers selected by previous queries.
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -187,7 +192,9 @@ in your terminal._
 You can specify the browser and Node.js versions by queries (case insensitive):
 
 * `defaults`: Browserslist’s default browsers
-  (`> 0.5%, last 2 versions, Firefox ESR, not dead`).
+  (`> 0.5%, last 2 versions, safe, not dead`).
+* `balanced`: target modern (updated) browsers (`last 2 versions, safe, not
+  unsafe, not < 0.1%`).
```
ai, line 197: I against removing non-popular browsers (`not < 0.1%`), because it will block new browsers coming to the market (nobody supports it because nobody use them / nobody uses them because nobody supports them).

index.js
```diff
@@ -1114,6 +1126,30 @@ var QUERIES = [
...
+      ]
+      return resolve(safe, context)
+    }
+  },
+  {
+    // In `last 2 versions` but doesn't receive _security_ updates.
+    regexp: /^unsafe$/i,
+    select: function (context) {
+      var unsafe = [
+        'ie 11', // No more security fixes
```
ai, line 1146: Any proofs that IE does not receive security updates?

README.md
```diff
@@ -222,6 +229,10 @@ You can specify the browser and Node.js versions by queries (case insensitive):
 * `dead`: browsers without official support or updates for 24 months.
   Right now it is `IE 10`, `IE_Mob 10`, `BlackBerry 10`, `BlackBerry 7`,
   `Samsung 4` and `OperaMobile 12.1`.
+* `safe`: browsers that are not part of `last 2 versions` anymore but that are
+  still receiving security patches.
+* `unsafe`: browsers that are still part of `last 2 versions` but that are not
```
ai, line 234: The definition of this query is equal to `dead`

index.js
```diff
@@ -1114,6 +1126,30 @@ var QUERIES = [
...
     }
   },
+  {
+    // Not in `last 2 versions` but still receive _security_ updates.
+    regexp: /^safe$/i,
+    select: function (context) {
+      var safe = [
+        'firefox esr', // Long-Term-Support
+        'ios_saf 12', // Still active
+        'edge 18' // Still active
```
ai, line 1136: We can add `edge 18` to the `default`.

Any link about iOS 12 support?

MisterTicot:
@ai Thanks for the quick review. I think you'll find the answers for your question on the [analysis](https://github.com/browserslist/browserslist/issues/435#issuecomment-575407090) I linked and that I took hours to put together. If not, please let me now about it.

ai:
I found this link about IE, but it does not tell, that IE has no security updates
https://www.zdnet.com/article/microsoft-security-chief-ie-is-not-a-browser-so-stop-using-it-as-your-default/

It tells that it is not recommended, not about security updates.

---

I do not found any good reasons to remove non-popular browsers there.

README.md
MisterTicot: Please note that it only blocks browsers when they _ don't respect the standards_. From a webapp developer point of view, overloading each & every visitors with dozens if not hundreds Kos of polyfills only to support a browser that has under 0.1% market share & a low standard compliance is a big no-no - especially as bandwidth is a also an accessibility matter.

ai: > Please note that it only blocks browsers when they _ don't respect the standards_.

Why? It is the last query, it takes browsers from the right (`last 2 versions, safe, not unsafe`) and removes all `< 0.1%` browsers from there.

It will remove all non-popular browsers even if they support standards.

MisterTicot: (more generally, as explained in the link, the point of that proposal is to achieve a saner polyfills set by cutting off browsers that are expensive to support but that may not be part of sophisticated webapps targetted audience)

ai: > may not be part of sophisticated webapps targetted audience

Browserslist is about keeping diversity. I am from Russia. We all used the Opera 10 years ago. But nobody used Opera outside of Russia. And most of the websites did not support us.

I created Browserslist to avoid removing minor browsers just because Western developers do not know about them.

I understand that you want to re-think the default browser query and created a good alternative. Splitting the audience to “sophisticated” and non-sophisticated will not work as argument for me. We must keep browsers popular in Afrika and China.

MisterTicot: > It will remove all non-popular browsers even if they support standards.

The fact that a browser is not in the list doesn't means that it won't be supported. For example, if your list includes Chrome 80 then chances are Chrome 70+ will all work because they have the same level of support for the standards. Likewise, any browser following the standards will work.

Practically, the fact that a browser is not included in the list means that the webapp won't include polyfills that are required only by that browser.

ai: Browserslist is also used in `Your browser is not supported` banners.

Let’s speak honestly, what browsers exactly do you want to remove by this query?

MisterTicot: I just said it - anything that has a very low audience but require dozens of Kos of polyfills. That's a dynamic rule, that's not aimed at a brand in particular.

ai: 1. `not < 0.1%` != “a very low audience but require dozens of Kos of polyfills”, because you remove browsers with good spec support
2. I do not think that it is a good ethical choice. I do not want to block minorities from the Internet. In the `not dead` we at least promoting Internet security.

Suggest different criteria.

PR #479, opened 2020-04-21
Upgrade `electron-to-chromium`

veado:
to support latest Electron `v.8.2.x`.


veado:
Oops, `browserslist` has been updated over-night. So this fix is not needed anymore. Thx!

_Closing now ..._

Issue #486, opened 2020-05-21
Upgrade electron-to-chromium to 1.3.448 because the current used version does not contain Electron 8.3.0


kotapeter:
If I use browserlist in Electron 8.3.0, I get the following error:

```
BrowserslistError: Unknown version 8.3.0 of electron
```

Solution:
Upgrade `electron-to-chromium` package to 1.3.448. It contains this electron version.

https://github.com/Kilian/electron-to-chromium/blob/v1.3.448/full-versions.js

Thank you!

ai:
Do you want send a PR and update error message to suggest using dependencies?

kotapeter:
@ai yes, It would be great:) Do I need any permission to create a pr?

kotapeter:
@ai sure, I've opened a pr:)

PR #487, opened 2020-05-21
chore: update electron-to-chromium dependency

kotapeter:
https://github.com/browserslist/browserslist/issues/486


ai:
You do not need to update dependency here. Just run `npm update` and npm will update all dependencies (or remove and then install Browserslist again).

My suggestion was to change error message on unknown version

PR #497, opened 2020-06-19
chore: simplify `package.json` searcher

lukeed:
The `pkg-up` makes up more than have of `browserslist`'s dependencies ([graph](https://npm.anvaka.com/#!/view/2d/browserslist)). It's pretty deceiving, actually.

Instead, `escalade` has zero dependencies, is more flexible, and is also faster ([results](https://github.com/lukeed/escalade#benchmarks)).
The `escalade/sync` version supports Node 6.x and above.

Finally, `pkg-up` (because of `find-up`) can be seen as a security issue since it traverses all the way up to your system's root directory (`/`). With `escalade`, traversal stops at the `process.cwd()` location.

### DIFF
--- a/test/update-db.test.js
+++ b/test/update-db.test.js
@@ -1,7 +1,7 @@
 let { remove, copy, readFile, ensureDir } = require('fs-extra')
 let { execSync } = require('child_process')
 let { nanoid } = require('nanoid/non-secure')
-let { homedir } = require('os')
+let { tmpdir } = require('os')
 let { join } = require('path')
 
 let updateDd = require('../update-db')
@@ -17,7 +17,7 @@ afterEach(async () => {
 })
 
 async function chdir (fixture, ...files) {
-  testdir = join(homedir(), `browserslist-test-${ fixture }-${ nanoid() }`)
+  testdir = join(tmpdir(), `browserslist-${ fixture }-${ nanoid() }`)
   await ensureDir(testdir)
 
   let from = join(__dirname, 'fixtures', fixture)
--- a/test/update-db.test.js
+++ b/test/update-db.test.js
@@ -1,7 +1,7 @@
 let { remove, copy, readFile, ensureDir } = require('fs-extra')
 let { execSync } = require('child_process')
 let { nanoid } = require('nanoid/non-secure')
-let { tmpdir, homedir } = require('os')
+let { homedir } = require('os')
 let { join } = require('path')
 
 let updateDd = require('../update-db')
@@ -17,7 +17,7 @@ afterEach(async () => {
 })
 
 async function chdir (fixture, ...files) {
-  testdir = join(homedir(), tmpdir(), `browserslist-${ fixture }-${ nanoid() }`)
+  testdir = join(homedir(), `browserslist-test-${ fixture }-${ nanoid() }`)
   await ensureDir(testdir)
 
   let from = join(__dirname, 'fixtures', fixture)
--- a/test/update-db.test.js
+++ b/test/update-db.test.js
@@ -1,7 +1,7 @@
 let { remove, copy, readFile, ensureDir } = require('fs-extra')
 let { execSync } = require('child_process')
 let { nanoid } = require('nanoid/non-secure')
-let { tmpdir } = require('os')
+let { tmpdir, homedir } = require('os')
 let { join } = require('path')
 
 let updateDd = require('../update-db')
@@ -17,7 +17,7 @@ afterEach(async () => {
 })
 
 async function chdir (fixture, ...files) {
-  testdir = join(tmpdir(), `browserslist-${ fixture }-${ nanoid() }`)
+  testdir = join(homedir(), tmpdir(), `browserslist-${ fixture }-${ nanoid() }`)
   await ensureDir(testdir)
 
   let from = join(__dirname, 'fixtures', fixture)
--- a/update-db.js
+++ b/update-db.js
@@ -5,11 +5,11 @@ var fs = require('fs')
 
 var BrowserslistError = require('./error')
 
-// eslint-disable-next-line consistent-return
 function filterPkg (dir, names) {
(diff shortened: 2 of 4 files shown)


lukeed:
This change, along with the others made to Autoprefixer and PostCSS recently, will mean that standard Autoprefixer + Browserslist + PostCSS trio drops **15** dependencies out of the [current 25 mesh dependencies](https://npm.anvaka.com/#!/view/2d/autoprefixer/9.8.0)~!

ai:
> Finally, pkg-up (because of find-up) can be seen as a security issue since it traverses all the way up to your system's root directory (/). With escalade, traversal stops at the process.cwd() location.

It will be a breaking changes.

Also I have use case to not stop at `cwd`:

```sh
$ ls
frontend/ admin-frontend/ backend/ package.json
$ cd frontend/
$ npm start
```

On my project we have a separated npm projects with the common `package.json` at root.

Can we disable this behaviour and scan until the root or `$HONE`?

lukeed:
Ah, true, good point. 

Yes I can accomodate and search until [`os.homedir()`](https://nodejs.org/api/os.html#os_os_homedir). It will have to be a 2.0.
How does that sound?

ai:
Sounds good

test/update-db.test.js
```diff
@@ -17,7 +17,7 @@ afterEach(async () => {
 })
 
 async function chdir (fixture, ...files) {
-  testdir = join(tmpdir(), `browserslist-${ fixture }-${ nanoid() }`)
+  testdir = join(homedir(), tmpdir(), `browserslist-${ fixture }-${ nanoid() }`)
```
lukeed, line 20: I had to do this since `escalade` relies on matching `os.homedir()` to continue

test/update-db.test.js
```diff
@@ -17,7 +17,7 @@ afterEach(async () => {
 })
 
 async function chdir (fixture, ...files) {
-  testdir = join(tmpdir(), `browserslist-${ fixture }-${ nanoid() }`)
+  testdir = join(homedir(), `browserslist-test-${ fixture }-${ nanoid() }`)
```
lukeed, line 20: I had to do this since `escalade` relies on matching `os.homedir()` to continue searching.

update-db.js
```diff
@@ -1,12 +1,19 @@
 var childProcess = require('child_process')
-var pkgUp = require('pkg-up')
+var escalade = require('escalade/sync')
 var path = require('path')
 var fs = require('fs')
 
 var BrowserslistError = require('./error')
 
+// eslint-disable-next-line consistent-return
```
ai, line 8: I prefer to avoid disabling ESLint. Do we have any other options?

test/update-db.test.js
ai: We can’t allow test creating files in user’s home directory

update-db.js
lukeed: The callback signature is `string | falsey` -- the only way to satisfy ESLint will be to return an empty string for false condition. Is that fine with you?

test/update-db.test.js
lukeed: The only other option I can think of is to create temporary directories within the `test/fixtures` location.

update-db.js
ai: > Is that fine with you?

yeap

test/update-db.test.js
ai: What will be extact problems of calling `escalade` inside the `/tmp`?

lukeed: Not to worry – you convinced me. 
Many other projects are going to expect the exact same behavior as `find-up`, so I'm just going to match it. I didn't originally agree with it, but if I'm going to share this module I might as well accommodate others' current needs.

`escalade@2.0` looped so long as `dir.startsWith(homedir)`. So if `os.homedir()` is `/Users/lukeed` then `/tmp/var` fails instantly.

I'm releasing escalade@3.0 right now 😂 I tested it locally and it works with the original test code.

lukeed:
Thanks for your patience & reviews on this :)

ai:
Sorry. Have no time to look again because of the work.

Will try to review tomorrow or aftertomorrow.

lukeed:
No rush! Whenever is best for you 👍

ai:
Thanks. I [simplified code](https://github.com/browserslist/browserslist/commit/1c5209569cea5908298b39581d5ceb05f9c8caa1) a little and now it looks even better than before. I also [limited `engines`](https://github.com/browserslist/browserslist/commit/f706fe4b3e3f8e67dee0316351bc2ee9c1039eef) to avoid Node.js 11 bug here as well.

I will release this pull request tomorrow.

---

I will use this good moment to talk about a bad thing. I decided that switch to [`colorette`](https://www.npmjs.com/package/colorette) because their API fits my needs a little better.

But you gave amazing feedback in your projects and working with you is reallyt great.

lukeed:
Cool :)

Ah, that's okay – I just wanted to get rid of chalk. What was the difference in API btw? It looks the same.
I had meetings this afternoon, but am working on tests for your TTY PR now.

ai:
> What was the difference in API btw? It looks the same.

1. One file instead of separated.
2. I personally do not like `kleur.$.enabled` compare to `options.enabled`. But it is very subjective.
3. I like color detection (outside of `isTTY`) a little more (but it is again subjective)

lukeed:
Ok 👍 Again, I'm just glad I'm not downloading 7 dependencies to do one thing. I guess two is better (my tools all use kleur).

1. `kleur/colors` is a separate module entirely. This was better than releasing a second, similar package. User picks the style they like
2. Yeah, minor :) It had to be that was because of ESM -- readonly exports.
3. Seems to only add `NO_COLOR`. Windows check is not needed anymore since Node 2.x or something.

ai:
Released in 4.12.1

PR #499, opened 2020-06-24
Add explanation of and/or combinations in a human language.

sznowicki:
Hello everyone and first of all many  thanks for inventing and maintaining this. Huge fan.

I was living in a wrong state of understanding how the queries work for some time. 

Today I started to dig into this topic while fighting with babel to stop transpiling `async..await` statements and after reading the README I started to be even more confused.

This PR already consists of small suggestion how to explain `and/or` combinations better but I treat this more like a starting point for further explanations (thanks in advance) or maybe even discussion whether there's a missing feature/bug that causes confusion.

## 1. Do I read it right?
So, as I understand from the README, every "and" statement applies to a previous query. 

Therefore this query:
`chrome > 75 or ie > 9 and last 1 version` should evaluate into:
`chrome > 75 or ie > 10`.

What I see both in the `npx browserslist` and on the website, this doesn't seem to be true:

Picture1: website screenshot:
Screenshot 2020-06-24 at 22 01 31

Picture2: IDE screenshot:
Screenshot 2020-06-24 at 22 06 04

This is addressed directly in my PR by changing "previous query" into "all previous queries". Which seems to be what's actually happens.

Picture 3: website screenshot with moved `and` condition:
Screenshot 2020-06-24 at 22 01 45

## 2. package.json confusion

This discovery leads to a probable bigger issue with a package.json configuration.

This configuration (with having the above knowledge) would feel natural for me:
```json
"browserslist": [
    "chrome > 75",
    "ie > 9",
    "and last 1 version"
  ],
```

This however ends up with an error:
`browserslist: Unknown browser query `and last 1 version`. Maybe you are using old Browserslist or made typo in query.`

If I change it to a valid one:
```json
"browserslist": [
    "chrome > 75",
    "ie > 9 and last 1 version"
  ],
```

it's well, valid but without the knowledge above and given the "previous query" statement in the documentation, leads to unexpected behaviour evaluating to `(chrome > 75 or ie > 9) AND last 1 version`.

I don't really know if what I see is a bug or a future but the configuration above makes me think it's `chrome or (ie >9 and last 1 version)` because of the array notation and the fact that last two queries are bundled together. 

This however might be only my perception.

I'm looking forward hearing from you and again thanks for everything.

### DIFF
--- a/README.md
+++ b/README.md
@@ -165,8 +165,8 @@ An `or` combiner can use the keyword `or` as well as `,`.
 
 `and` query combinations are also supported to perform an
 intersection of all the previous queries: `last 1 version or chrome > 75 and > 1%`,
-which in human words matches:
-`last 1 version of any browser with more than 1% of the market share and every version of chrome higher than 76 with more than 1% market share`.
+which in pseudo code matches:
+`match = (is browser last version OR is chrome since v76) &&  has more than 1% marketshare`
 
 There is 3 different ways to combine queries as depicted below. First you start
 with a single query and then we combine the queries to get our final list.
--- a/README.md
+++ b/README.md
@@ -164,7 +164,9 @@ An `or` combiner can use the keyword `or` as well as `,`.
 `last 1 version or > 1%` is equal to `last 1 version, > 1%`.
 
 `and` query combinations are also supported to perform an
-intersection of the previous query: `last 1 version and > 1%`.
+intersection of all the previous queries: `last 1 version or chrome > 75 and > 1%`,
+which in human words matches:
+`last 1 version of any browser with more than 1% of the market share and every version of chrome higher than 76 with more than 1% market share`.
 
 There is 3 different ways to combine queries as depicted below. First you start
 with a single query and then we combine the queries to get our final list.


ai:
Oh. Too long issue for quick response 😅.

I will fix other issues and will return to this one later.

ai:
> So, as I understand from the README, every "and" statement applies to a previous query.

Nope. It applies to all previous queries. So `chrome > 75 or ie > 9 and last 1 version` it like `(chrome > 75 or ie > 9)  and last 1 version`, which follow us to `chrome 78 or ie 11`.

> This configuration (with having the above knowledge) would feel natural for me

It feels natural for me too,  but we do not support it.

Honestly, I think `and` is too math-y syntax and it will be hard for you later to understand your query. I prefer the simple `,` (`or`).

But I am OK if you will send `and X` query support if it will not reduce code maintainability.

README.md
```diff
@@ -164,7 +164,9 @@ An `or` combiner can use the keyword `or` as well as `,`.
 `last 1 version or > 1%` is equal to `last 1 version, > 1%`.
 
 `and` query combinations are also supported to perform an
-intersection of the previous query: `last 1 version and > 1%`.
+intersection of all the previous queries: `last 1 version or chrome > 75 and > 1%`,
+which in human words matches:
+`last 1 version of any browser with more than 1% of the market share and every version of chrome higher than 76 with more than 1% market share`.
```
ai, line 169: Multiline queries do not look beautiful for me



Can we use a quote and simplify the example, because it is hard to understand so long sentence.

ai:
BTW, I like this doc additional. We definitely should clarify that `and` applies to all previous queries.

README.md
sznowicki: Changed it to pseudocode. I think this would make more sense for developers and dev related people who are the audience.

```
which in pseudo code matches:
`match = (is browser last version OR is chrome since v76) &&  has more than 1% marketshare`
```

piranna:
> BTW, I like this doc additional. We definitely should clarify that `and` applies to all previous queries.

This clarify the behaviour, but intuitively, I think a more math-ly behaviour of `and` and `or` would make sense, but that would be a breaking change... :-/

Issue #491, opened 2020-05-28
Search by features


piranna:
Since we are already using [caniuse-lite](https://github.com/ben-eb/caniuse-lite) package that allow to search [by feature](https://github.com/ben-eb/caniuse-lite#litefeaturejs), add the option to do, similar to the recomendation of testing for features in browser instead of checking compatibility using the UserAgent string.

ai:
Dublicate https://github.com/browserslist/browserslist/issues/407

ai:
Fixed by @piranna in https://github.com/browserslist/browserslist/pull/501 and released in 4.13

Issue #407, opened 2019-08-27
Feature-based query


alaricd:
While this may not be an issue, I can't find a mailing list, or any example.

I want to limit the browsers I support to those that support tls12, caniuse has the info, but I can't find any way to query for that.

If it is a supported query, can we add documentation for that?

ai:
Nope, docs will be too big.

piranna:
For a browser library it would make sense, but for a development tool, how much does it matter if it gets increased? I don't think it would make the package so much bigger, and would allow to do more fine-grained optimization with tools like Babel or webpack that would have a real impact in the browser deployed code.

ai:
There are few use cases when development tool is using in the browser.

However, I have an another idea: you can create a build script, which will make sharable config for each Can I Use feature and publish these configs on npm. It will also be more flexible for tricky cases of partial feature support.

Creating these sharable configs is one of the goals on [our OpenCollective funding page](opencollective.com/browserslist).

alaricd:
Hi, 

I have created a tool using 

https://www.npmjs.com/package/browserslist 

and caniuse-api (https://github.com/Nyalab/caniuse-api) 

to create a tool that would quiery what I need. If I had more idea of how to bundle it, pass parameters etc, it could be used for any feature. Obviously my TLS query makes the most sense, but I made it to be able query anything i the caniuse database. I might have lost the exact code but I can always recreate it. 


Alaric

piranna:
@alaricd, have you published your code somewhere? Maybe I can do the final fixes.
@ai, would you accept a pull-request if it show to be not such a huge change?

alaricd:
No I haven't, i lost my hard drive after getting it working. I will take some time and get it recovered and working again and let you know. 

Sent from MailDroid

-----Original Message-----

piranna:
> No I haven't, i lost my hard drive after getting it working. I will take some time and get it recovered and working again and let you know.

Ok, don't worry, take your time :-)

ai:
Fixed by @piranna in https://github.com/browserslist/browserslist/pull/501 and released in 4.13

PR #501, opened 2020-06-30
Add support for features selection

piranna:
This pull-request add support for a `feature` query, allowing to ask for browsers that implement a particular feature, fixing #407 and #491 issues. Currently only checks if it has support at any state, , but could be possible to extend it for more fine-grain queries (non prefixed, etc).

It is a really surprising simple pull-request, so it would almost no affect package size since data and query functions are already there, it's just a matter of reading it. At first I added a `loadFeature()` function based on `loadCountry()` one, but later it showed that code was not being covered and seems data was already set by `caniuse-lite` library itself, so only the query function needed to be implemented. I have code on my machine, if this gives problems on CI, I can add it easily.


piranna:
So... tests are passing in CI server, definitely the `loadFeature()` function is not needed and features data was already being loaded in memory by `caniuse-lite` library itself :-)

ai:
@piranna I can’t accept this PR without a clear idea of how `loadFeature()` will be implemented without bloating JS bundle size

piranna:
> @piranna I can’t accept this PR without a clear idea of how `loadFeature()` will be implemented without bloating JS bundle size

As I told you, seems it's not needed, but in any case, this is the implementation of `loadFeature()` I was going to push:

```js
  loadFeature: function loadFeature (usage, code) {
    code = code.replace(/[^\w-]/g, '')
    if (!usage[code]) {
      // eslint-disable-next-line security/detect-non-literal-require
      var compressed = require('caniuse-lite/data/features/' + code + '.js')
      var usageData = feature(compressed).stats
      usage[code] = { }
      for (var i in usageData) {
        for (var j in usageData[i]) {
          usage[code][i + ' ' + j] = usageData[i][j]
        }
      }
    }
  },
```

Although the tests at `feature.test.js` were passing, lines inside the `if` conditional were not set, that's how I found that info was already provided by `caniuse-lite` at the `usage` argument and it was not needed to add the `loadFeature()` function at all. From what I've investigated, it's populated when browsers info is loaded.

piranna:
> it's populated when browsers info is loaded

Not sure if it's by design of `caniuse-lite` and so `browserlist` code can be simplified (for example, removing the `loadCountry()`), or it's a bug in the tests so they are not isolated, thought...

ai:
> As I told you, seems it's not needed

Sorry, I have too much work right now and could miss something in open source discussions. Where do you mention it?

> require('caniuse-lite/data/features/' + code + '.js')

This code will force webpack to put all files inside `caniuse-lite/data/features/` to the bundle of Browserslist, which is a huge size.

browser.js
```diff
@@ -21,7 +21,7 @@ module.exports = {
 
   loadCountry: function loadCountry () {
     throw new BrowserslistError(
-      'Country statistics is not supported ' +
+      'Country statistics are not supported ' +
```
ai, line 24: Can you move this to another PR?

piranna:
> > As I told you, seems it's not needed
> 
> Sorry, I have too much work right now and could miss something in open source discussions. Where do you mention it?

In the [previous comment](https://github.com/browserslist/browserslist/pull/501#issuecomment-652110587): 

```
So... tests are passing in CI server, definitely the loadFeature() function is not needed and features data was already being loaded in memory by caniuse-lite library itself :-)
```


> > require('caniuse-lite/data/features/' + code + '.js')
> 
> This code will force webpack to put all files inside `caniuse-lite/data/features/` to the bundle of Browserslist, which is a huge size.

`size-limit` check is passing, bundle size is 12.2KB while the top limit is at 13KB... I think they are not included at all because it's being used the `browser.js` file instead of the `node.js` one.

ai:
> So... tests are passing in CI server, definitely the loadFeature() function is not needed and features data was already being loaded in memory by caniuse-lite library itself :-)

I did understand it. CI tests are passing because you mocked `loadFeature` in the tests only.

> size-limit check is passing, bundle size is 12.2KB while the top limit is at 13KB... I think they are not included at all because it's being used the browser.js file instead of the node.js one.

You didn’t add `loadFeature` implementation to `node.js` or `browser.js`.

package.json
```diff
@@ -49,8 +49,8 @@
     "eslint-plugin-unicorn": "^16.1.1",
     "fs-extra": "^8.1.0",
     "husky": "^4.2.5",
-    "jest": "^25.4.0",
     "jest-ci": "^0.1.1",
+    "jest-cli": "^25.5.4",
```
ai, line 53: Why did you add this change?

piranna:
> CI tests are passing because you mocked `loadFeature` in the tests only

> You didn’t add loadFeature implementation to node.js or browser.js

I've not mocked nor added it anywhere, because it's not needed at all, `caniuse-lite` already has the info set... In case you are curious, the browser version of `loadFeature()` I was going to push was:

```js
  loadFeature: function loadFeature () {
    throw new BrowserslistError(
      'Feature statistics are not supported ' +
      'in client-side build of Browserslist')
  },
```

Just the same behaviour of browser version of `loadCountry()`.

package.json
piranna: Tests were not passing me, not sure if by using `npm` instead of `yarn`. They gave me an error of not being able to find a file inside `jest-cli` package, probably because you have `jest-cli` installed globally. I've check that in the dependencies is being used `jest`, while it's a placeholder package and the recommended one is in fact `jest-cli`, that's why I've changed it.

browser.js
piranna: Sure, done at #502 :-) I've already removed the change from this pull-request, too.

ai:
> I've not mocked nor added it anywhere

You mocked them here https://github.com/browserslist/browserslist/pull/501/files#diff-ec67b11dadaf98465d1ed79041d11266R6

Try to run this command in your branch:

```
./cli.js "feature rtcpeerconnection"
```

It will not work because you do not have a Node.js’s implementation of `loadFeature` in the branch.

package.json
ai: > probably because you have jest-cli installed globally

Nope. I do not install npm package globally (I think it is a very bad practice)

piranna:
> You mocked them here https://github.com/browserslist/browserslist/pull/501/files#diff-ec67b11dadaf98465d1ed79041d11266R6

Ok, my fault... It was not that `caniuse-lite` was already filling the info, it was mocked... that's the reason why code was not covered :-( I'm going to restore the code and add a test without mock, and push the changes again, thank for the catch up :-)

piranna:
I've just added the actual implementations of `loadFeature()` sorry for misunderstood the mock. Tests are passing, and I've checked the `size-limit` and it's 12.15 KB, so the features info is not being included in the browser version.

package.json
piranna: Commands were recommended to be installed globally, but except for particular use cases, yes, now it's recommended to install them locally.

README.md
```diff
@@ -202,6 +202,7 @@ You can specify the browser and Node.js versions by queries (case insensitive):
 * `cover 99.5%`: most popular browsers that provide coverage.
 * `cover 99.5% in US`: same as above, with [two-letter country code].
 * `cover 99.5% in my stats`: uses [custom usage data].
+* `feature rtcpeerconnection`: browsers with support for specific features.
```
ai, line 205: The new query do not follow “simple English sentence” look of other queries. `supports X` can be better.

ai: Also we need to clarify how to get feature name

index.js
```diff
@@ -887,6 +887,19 @@ var QUERIES = [
       return result
     }
   },
+  {
+    regexp: /^feature\s+(\w+)$/,
+    select: function (context, feature) {
+      env.loadFeature(browserslist.usage, feature)
+      var usage = browserslist.usage[feature]
+      return Object.keys(usage).reduce(function (result, version) {
+        if (usage[version] !== 'n') {
```
ai, line 896: What if it will be `n #1` (not, with a comment)?

node.js
```diff
@@ -237,6 +238,21 @@ module.exports = {
     }
   },
 
+  loadFeature: function loadFeature (usage, code) {
+    code = code.replace(/[^\w-]/g, '')
+    if (!usage[code]) {
+      // eslint-disable-next-line security/detect-non-literal-require
+      var compressed = require('caniuse-lite/data/features/' + code + '.js')
+      var usageData = feature(compressed).stats
+      usage[code] = { }
```
ai, line 247: `usage` cache was named for usage statistics. We need to rename it if we want to use it for feature cache.

README.md
piranna: > The new query do not follow “simple English sentence” look of other queries. `supports X` can be better.

Didn't know about project was using the "simple English sentence" concept, fixed.


> Also we need to clarify how to get feature name

Done, I've put a link similar to the regions one.

node.js
piranna: Not fully sure if I understand it, it's a local variable... Are you talking about `browserslist.usage`? Ok, make sense to use a differente cache. What's a better name, `browserslist.features` or `browserslist.supports`?

index.js
piranna: It would fail :-) I did just a simple implementation with basic support, correct one would need to actually parse the info fields, I'm working on that.

piranna: Finally I've just changed the conditional to use `indexOf()` instead of a strict checking for `n`. Where can I find info about the meaning of all the flags that can be set? They are defined at https://github.com/ben-eb/caniuse-lite/blob/master/src/lib/supported.js but don't know the meaning of all of them... :-/ With that info, I could change the regexp so it could be possible to match `supports partially XXX` or `supports not deprecated YYY`, for example.

piranna: https://github.com/Fyrd/caniuse/blob/master/CONTRIBUTING.md#supported-changes

Found it! :-)

> stats — The collection of support data for a given set of browsers/versions. Only the support value strings can be modified; additional versions cannot be added. Values are space-separated characters with these meanings, and must answer the question "Can I use the feature by default?":
y - (Y)es, supported by default
a - (A)lmost supported (aka Partial support)
n - (N)o support, or disabled by default
p - No support, but has (P)olyfill
u - Support (u)nknown
x - Requires prefi(x) to work
d - (D)isabled by default (need to enable flag or something)
#n - Where n is a number, starting with 1, corresponds to the notes_by_num note. For example: "42":"y #1" means version 42 is supported by default and see note 1.

ai: I suggest:

```js
let support = usage[version][0]
if (support !== 'n' && support !== 'p' && support !== 'd')
```

node.js
ai: Yeap, I do not like that name of local variabe `usage` because it doesn’t explain the content in the variable. Seems like it was copied from `loadCountry` when it was related to “usage statistics”.

README.md
```diff
@@ -232,6 +234,7 @@ You can specify the browser and Node.js versions by queries (case insensitive):
 You can add `not ` to any query.
 
 [`caniuse-lite/data/regions`]: https://github.com/ben-eb/caniuse-lite/tree/master/data/regions
+[`caniuse-lite/data/features`]: https://github.com/ben-eb/caniuse-lite/tree/master/data/features
```
ai, line 237: It is hard to use this dir because you can’t read the description in those files.

It will be better to show that it is a `feat` parameter on Can I Use page.

browser.js
```diff
@@ -25,6 +25,12 @@ module.exports = {
       'in client-side build of Browserslist')
   },
 
+  loadFeature: function loadFeature () {
+    throw new BrowserslistError(
+      'Feature statistics are not supported ' +
```
ai, line 30: `statistics`is a wrong word copied from `loadCountry`. `supports query is not …`

index.js
```diff
@@ -887,6 +887,19 @@ var QUERIES = [
       return result
     }
   },
+  {
+    regexp: /^supports\s+(\w+)$/,
```
ai, line 891: `-` is not part of `\w` but widely used in `feat` code

index.js
```diff
@@ -887,6 +887,19 @@ var QUERIES = [
...
+    select: function (context, feature) {
+      env.loadFeature(browserslist.usage, feature)
+      var usage = browserslist.usage[feature]
```
ai, line 894: let’s rename `browserslist.usage` too to `browserslist.cache` and clean it in `browserslist.clearCaches`

test/feature.test.js
```diff
@@ -0,0 +1,29 @@
...
+
+it('selects browsers by feature', () => {
+  browserslist.usage = {
+    rtcpeerconnection: {
+      'and_chr 81': 'y',
+      'firefox 2': 'n'
+    }
+  }
+
+  expect(browserslist('supports rtcpeerconnection')).toEqual(['and_chr 81'])
```
ai, line 28: let’s test feature with `-` in the name

index.js
piranna: I've reconsider this, at first I wanted "support, no matter how", but your proposal gave the idea that's better to filter for "native support provided by the browser itself, without intervention from developer or user". That's lead us only to full and partial support, both with prefix or not, so I've changed the conditional to `flags.indexOf('y') >= 0 || flags.indexOf('a') >= 0` (I've inverted the conditional since this way was smaller, with the new conditions).

node.js
piranna: Done, I've renamed it to `features`.

README.md
piranna: Fixed, I've added an explanation of the `feat` parameter, and left the link as reference.

browser.js
piranna: Good point, fixed.

index.js
piranna: Added support for dashes on features names.

index.js
piranna: Done.

test/feature.test.js
piranna: Done.

README.md
```diff
@@ -202,6 +202,9 @@ You can specify the browser and Node.js versions by queries (case insensitive):
 * `cover 99.5%`: most popular browsers that provide coverage.
 * `cover 99.5% in US`: same as above, with [two-letter country code].
 * `cover 99.5% in my stats`: uses [custom usage data].
+* `supports rtcpeerconnection`: browsers with support for specific features.
+  `rtcpeerconnection` here is the `feat` parameter at the URL of the [Can I Use]page. A list of all available features can be found at
```
ai, line 206:
```suggestion
  `rtcpeerconnection` here is the `feat` parameter at the URL of the [Can I Use] page. A list of all available features can be found at
```

ai:
Great work. Thanks. Ths last fix and we are ready to merge and release it.

What is your Twitter name to give you a credit in release tweet of [Browserslist twitter]()?

README.md
piranna: Good catch, I've missed it just in the new line wrapping :+1:, fixed.

piranna:
> Great work. Thanks. Ths last fix and we are ready to merge and release it.
> 
> What is your Twitter name to give you a credit in release tweet of [Browserslist twitter](https://twitter.com/browserslist)?

You are welcome :-) My twitter account is https://twitter.com/el_piranna. Can you also be able to add a reference to my client [CPI Technologies GmbH](https://cpitech.io) or its web app [Veedeo.me](http://veedeo.me/)? They have been sponsoring me to add this functionality as part of my work with them :-)

ai:
Released in 4.13

piranna:
> Released in 4.13

Great, thank you! :-D

PR #509, opened 2020-07-19
Add node version 13.2 in the package.json list

mshafiq235:


ai:
1. What is the reason of this change?
2. Will it prevent using Browserslist with 13.5-13.6 which have ES modules bug?
3. Are you sure that 13.2 does not have ES module bug?

mshafiq235:
@ai My bad, I wasn't aware of that bug. I have been using it for couple of projects and it seems to be working fine. I think I will have to update to 13.7 since my project one of my project dependency has dependency on Browserslist, which fails due to node 13.2.

ai:
Do not update to 13.7! It doesn't have security updates.

It was a short-term release, which shouldn't be used on production.

You need to update to long-term release like 14.x.

PR #511, opened 2020-07-23
feat: add `normalizers` (byEngine and toDesktop)

### LINKED ISSUE #508: Option to return Chrome versions for Chromium-based browsers
Due to insufficient compatibility data, most tools ignore browsers unpopular in US, such as UC Browser and QQ Browser. (e.g., https://github.com/babel/babel/pull/8552) But as far as I understand, this is the very reason Browserslist is created. Fortunately, most of them are based on Chromium and compatibility is almost identical to the version of Chromium they are based on. So an option to return corresponding Chrome versions for those browsers would be useful and encourage people to support them.

### LINKED ISSUE #510: `mobileToDesktop` doesn't apply to usage queries
```sh
npx browserslist --mobile-to-desktop ">3%"
```
**Actual**
```
and_chr 81
chrome 83
ios_saf 13.4-13.5
```

**Expected**
```
chrome 81
chrome 83
ios_saf 13.4-13.5
```

Currently, mobileToDesktop logic is only applied to `byName`:
https://github.com/browserslist/browserslist/blob/f5ff062d9ed723bfa7de9121ddeef27e1e7aa5fe/index.js#L245-L249

Is this intended behavior?

ylemkimon:
* docs: update `mobileToDesktop` documentation to match its behavior
* feat: add normalizers
* feat: add `byEngine` and `toDesktop` normalizers

Fixes #508, fixes #510.

### DIFF
--- a/README.md
+++ b/README.md
@@ -522,6 +522,13 @@ Options:
 * `normalizers`: An array of names of normalizers below and functions
   that accept a browser and return the normalized versions, which are
   applied to resolved browsers in order. For instance, `['byEngine', 'toDesktop']`.
+  * `byEngine`: Normalize Chromium-based browsers to Chrome. For instance,
+    UC Browser, QQ Browser, Baidu browser for Android, and Samsung Internet
+    will return `and_chr` with the version of Chromium they are based on.
+    Note Edge and Opera is not normalized. Gecko-based browsers are
+    also normalized to Firefox, e.g., KaiOS Browser will return `and_ff`.
+  * `toDesktop`: Normalize mobile browsers to desktop browsers.
+    For instance, Browserslist will return `chrome 20` on `and_chr 20`
 
 For non-JS environment and debug purpose you can use CLI tool:
 
--- a/README.md
+++ b/README.md
@@ -515,10 +515,10 @@ Options:
   Default is `false.`
 * `dangerousExtend`: Disable security checks for `extend` query.
   Default is `false.`
-* `mobileToDesktop`: Use desktop browsers if Can I Use doesn’t have data
-  about this mobile version. For instance, Browserslist will return
-  `chrome 20` on `and_chr 20` query (Can I Use has only data only about
-  latest versions of mobile browsers). Default is `false`.
+* `mobileToDesktop`: Resolve the mobile version using the desktop version when
+  Can I Use doesn't have data about the specified version. For instance, it
+  will return two Chrome for Android versions even if Can I Use has only
+  data only about the latest version. Default is `false`.
 
 For non-JS environment and debug purpose you can use CLI tool:
 
--- a/README.md
+++ b/README.md
@@ -519,6 +519,9 @@ Options:
   Can I Use doesn't have data about the specified version. For instance, it
   will return two Chrome for Android versions even if Can I Use has only
   data only about the latest version. Default is `false`.
+* `normalizers`: An array of names of normalizers below and functions
+  that accept a browser and return the normalized versions, which are
+  applied to resolved browsers in order. For instance, `['byEngine', 'toDesktop']`.
 
 For non-JS environment and debug purpose you can use CLI tool:
 
--- a/cli.js
+++ b/cli.js
@@ -19,6 +19,7 @@ var USAGE = 'Usage:\n' +
             '  npx browserslist --env="environment name defined in config"\n' +
             '  npx browserslist --stats="path/to/browserlist/stats/file"\n' +
             '  npx browserslist --mobile-to-desktop\n' +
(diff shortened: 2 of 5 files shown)


README.md
```diff
@@ -515,10 +515,21 @@ Options:
...
-  `chrome 20` on `and_chr 20` query (Can I Use has only data only about
-  latest versions of mobile browsers). Default is `false`.
+* `mobileToDesktop`: Resolve the mobile version using the desktop version when
+  Can I Use doesn't have data about the specified version. For instance, it
+  will return two Chrome for Android versions even if Can I Use has only
+  data only about the latest version. Default is `false`.
+* `normalizers`: An array of names of normalizers below and functions
+  that accept a browser and return the normalized versions, which are
+  applied to resolved browsers in order. For instance, `['toChromium', 'toDesktop']`.
+  * `toChromium`: Normalize Chromium-based browsers to Chrome. For instance,
```
ylemkimon, line 525: The name may be misleading as it also normalizes Gecko-based browsers to Firefox, but I cannot think of better name.

README.md
```diff
@@ -515,10 +515,21 @@ Options:
...
+    UC Browser, QQ Browser, Baidu browser for Android, and Samsung Internet
+    will return `and_chr` with the version of Chromium they are based on.
+    Note Edge and Opera is not normalized. Gecko-based browsers are
+    also normalized to Firefox, e.g., KaiOS Browser will return `and_ff`.
+  * `toDesktop`: Normalize mobile browsers to desktop browsers.
```
ylemkimon, line 530: I've named it `toDesktop` instead of `mobileToDesktop` as it may be confused with the existing option.

ai: What is the difference between `normalizers: ['toDesktop']` and `mobileToDesktop: true`?

README.md
ai: We can use `byEngine`

README.md
```diff
@@ -515,10 +515,21 @@ Options:
...
+    For instance, Browserslist will return `chrome 20` on `and_chr 20`
+* `keepOriginal`: Keep the original browser when normalizing with `normalizers`
```
ai, line 532: Can you explain why do we need this option?

README.md
ylemkimon: @ai See #510 and https://github.com/browserslist/browserslist/pull/511/commits/2ca9e82f71868a49473a80c2702b3b88793d9621. `mobileToDesktop` only resolves the version part using desktop data. For example:
```bash
$ node cli.js "last 5 and_chr versions"
and_chr 81
$ node cli.js --mobile-to-desktop "last 5 and_chr versions"
and_chr 83
and_chr 81
and_chr 80
and_chr 79
and_chr 78
$ node cli.js --normalizers=toDesktop "last 5 and_chr versions"
chrome 81
$ node cli.js --normalizers=toDesktop --mobile-to-desktop "last 5 and_chr versions"
chrome 83
chrome 81
chrome 80
chrome 79
chrome 78

```

README.md
ylemkimon: In case tools need or can process unnormalized data. I don't have any specific use case yet.

ai: Let’s remove it. I think the origin `mobileToDesktop` should be replaced.

README.md
ai: @JLHwung what do you think if we change `mobileToDesktop` behavior? On `last 5 and_chr versions` we will return `chrome 78-83` instead of `and_chr 78-83`. Will it affect Babel?

JLHwung: Internally Babel maps `and_chr` to `chrome` and it seems to me this PR moves this process to `browserslists`. I don't think it will affect Babel.

Another thoughts about `mobileToDesktop`: AFAIK opera_mob is mapped to opera, this is incorrect because they are if not always, frequently based on different chromium versions. Can we incorporate data from https://github.com/mdn/browser-compat-data/blob/master/browsers/opera_android.json ? This can help `preset-env` on supporting `opera_mob`. Surely we can address it in a different PR.

README.md
ylemkimon: @ai 
> I think the origin `mobileToDesktop` should be replaced.

I agree it'd be the best course of action, but it'd be a breaking change.

README.md
ylemkimon: I've added `'op_mob 46': 63` to `browserslist.normalizedVersions` (`byEngine`). I think removing it from `desktopNames` (`mobileToDesktop`) would be out of scope for this PR.

lastthyme (approved):

septatrix:
Would it be possible to also add support for [`QtWebEngine`](https://wiki.qt.io/QtWebEngine/ChromiumVersions) with the normalize logic introduced in the PR?

PR #513, opened 2020-07-24
feat: add `semver` and project version query to Node and Electron

### LINKED ISSUE #319: [Feature suggestion] Support `package.json` `engines.node` field
Suggestion for Node.js: default to `package.json` `engines.node` field (if present) instead of `last 2 versions`.

Use case: I am writing Node.js libraries using Node `11` syntax, but compiling with Babel in order to support Node `>=6`. I am declaring Node.js support with `engines.node`. Babel `preset-env` uses `browserslist`. Currently I need to duplicate the information from `engines.node` to a `browserslist` configuration file or `preset-env` configuration option.

### LINKED ISSUE #488: Difficult to select all versions in a major
If you have this in your `package.json`:

```json
{
  "engines": {
    "node": "^12.0.0 || >= 13.7.0"
  }
}
```

It is hard to replicate that in a browserslist query:

```
Node 12 - 13 and Node < 13, Node >= 13.7.0
```

One gotcha is that browserslist has a different behavior than the npm ecosystem when considering a major number on it's own: `Node 12` in browserslist will select only the newest v12 version, but `12` in `package.json` `engines` or `dependencies` selects every version in that major.

It would be nice if browserslist used the [`semver`](https://www.npmjs.com/package/semver) semantics, allowing the same semver version and range strings that we're used to using in `package.json`, e.g.

```
Node ^12.0.0 || >= 13.7.0
```

Or this, which is the same:

```
Node 12 || >= 13.7.0
```

ylemkimon:
* feat: add `semver` query to Node and Electron. Closes #488.
* feat: add project version query. Closes #319.

### DIFF
--- a/README.md
+++ b/README.md
@@ -215,9 +215,12 @@ You can specify the browser and Node.js versions by queries (case insensitive):
   * `maintained node versions`: all Node.js versions, which are [still maintained]
     by Node.js Foundation.
   * `node semver ^10.0.0`: all versions satisfying the [semver](https://www.npmjs.com/package/semver) range
+  * `project node`: all versions satisfying `engines.node` in package.json
 * `electron semver ^1.0.0`: all Electron versions satisfying the semver range.
   Note Browserslist has information only about major versions, so the patch
   version will be assumed as zero, i.e., >= 2.1.1 will not select 2.1.0.
+  * `project electron`: all versions satisfying `devDependencies.electron`
+    in package.json
 * `iOS 7`: the iOS browser version 7 directly.
   * `Firefox > 20`: versions of Firefox newer than 20.
     `>=`, `<` and `<=` work too. It also works with Node.js.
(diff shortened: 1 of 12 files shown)


ai:
What is use case?

I try to avoid adding more queries, because it is hard to maintain so many queries for me.

ylemkimon:
The major use case is deduplicating information in `package.json`, especially since Browserslist and semver have different version range semantics.

ai:
I do not like solving this problem by adding an extra query. We already have too many queries.

Maybe we need to change `node X` query behavior in a major version.

ai:
We will not be able to release major version in next month because I spending all my free time on preparing PostCSS 8.

ylemkimon:
@ai I've added `semver` query as it uses the same logic with the `project version` query, but if you think it causes too much technical debt, I can remove it.

ai:
I am right that you have no critical issue to be fixed with this PR?

Let me finish with PostCSS 8 release process and then I will think about Node.js versions.

ylemkimon:
@ai Yes, this is just a convenience.

chennxm (approved):

Issue #14, opened 2015-02-25
Can I target to all browsers?


lks2394758:
Say I'm targeting to all browsers, can I use something like browser: ['all']?

ai:
`> 0%`

For what case you need this?

rdbmax:
It can be usefull if you need to build an 'Update your browser page' with browser detection mechanism

dali-rajab:
you can also use `since 1990` 😃

PR #520, opened 2020-08-11
add example for all browsers to readme

tjenkinson:
I was looking for this. Thought it might be useful in the readme.

refs #14


ai:
I think the separated Examples section will be better for the cases like this.

In Examples we can also warn user about Babel/Autoprefixer slow performance and big bundle size.

ai:
You will need to add it and add a few other examples

pablopalacios:
Don't you think that `cover 100%` communicates better? It is also more clear if you are using custom stats: `cover 100% in my stats` (instead of `> 0% in my stats`).

PR #534, opened 2020-09-13
Updating node engines

itsdarrylnorris:
### Problem
I ran into a particular issue that my node version was 13.6 and it was throwing an error because how is declared in our engines. Updating to the 13.7 seems to fixed the issue.

### Solution
We can update the version to ^13 and include 14 since it's already out.

Thoughts? I have not  test this PR yet.


Semigradsky:
https://github.com/browserslist/browserslist/blob/master/CHANGELOG.md#4121

> Drop Node.js 13.0-13.6 support because of ES modules bug in that versions.

ai:
Browserslist do not work on Node.js < 13.7

PR #554, opened 2020-11-07
Implement target browser changes

AleksandrSl:
https://cultofmartians.com/tasks/browserslist-diff.html#task

### DIFF
--- a/test/fixtures/browserslist-diff/package.json
+++ b/test/fixtures/browserslist-diff/package.json
@@ -1,0 +1,16 @@
+{
+  "description": "browserslist",
+  "license": "MIT",
+  "repository": {},
+  "devDependencies": {
+    "autoprefixer": "^9.7.4",
+    "postcss-preset-env": "^6.7.0",
+    "qs": "^6.6.0"
+  },
+  "browserslist": [
+    "> 0.5%",
+    "last 2 versions",
+    "Firefox ESR",
+    "not dead"
+  ]
+}
--- a/test/update-db.test.js
+++ b/test/update-db.test.js
@@ -61,7 +61,7 @@ it('throws on missing lockfile', async () => {
 it('updates caniuse-lite for npm', async () => {
   let dir = await chdir('update-npm', 'package.json', 'package-lock.json')
 
-  expect(runUpdate()).toEqual(
+  expect(runUpdate()).toContain(
     'Current version: 1.0.30001030\n' +
     `New version: ${ caniuse.version }\n` +
     'Removing old caniuse-lite from lock file\n' +
@@ -79,7 +79,7 @@ it('updates caniuse-lite for npm', async () => {
 it('updates caniuse-lite without previous version', async () => {
   let dir = await chdir('update-missing', 'package.json', 'package-lock.json')
 
-  expect(runUpdate()).toEqual(
+  expect(runUpdate()).toContain(
     `New version: ${ caniuse.version }\n` +
     'Removing old caniuse-lite from lock file\n' +
     'Installing new caniuse-lite version\n' +
@@ -96,7 +96,7 @@ it('updates caniuse-lite without previous version', async () => {
 it('updates caniuse-lite for yarn', async () => {
   let dir = await chdir('update-yarn', 'package.json', 'yarn.lock')
 
-  expect(runUpdate()).toEqual(
+  expect(runUpdate()).toContain(
     'Current version: 1.0.30001035\n' +
     `New version: ${ caniuse.version }\n` +
     'Removing old caniuse-lite from lock file\n' +
@@ -118,7 +118,7 @@ if (!NODE_8 && (isInstalled('pnpm') || process.env.CI)) {
   it('updates caniuse-lite for pnpm', async () => {
     let dir = await chdir('update-pnpm', 'package.json', 'pnpm-lock.yaml')
 
-    expect(runUpdate()).toEqual(
+    expect(runUpdate()).toContain(
       'Current version: 1.0.30001035\n' +
       `New version: ${ caniuse.version }\n` +
       'Removing old caniuse-lite from lock file\n' +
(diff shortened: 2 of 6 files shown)


test/fixtures/browserslist-diff/package.json
```diff
@@ -0,0 +1,16 @@
+{
+  "description": "browserslist",
+  "license": "MIT",
+  "repository": {},
+  "devDependencies": {
+    "autoprefixer": "^9.7.4",
+    "postcss-preset-env": "^6.7.0",
+    "qs": "^6.6.0"
+  },
+  "browserslist": [
```
AleksandrSl, line 10: Specified intentionally to avoid effect of browserslist defaults on this test

test/update-db.test.js
```diff
@@ -133,3 +133,56 @@ if (!NODE_8 && (isInstalled('pnpm') || process.env.CI)) {
...
+    `New version: ${ caniuse.version }\n` +
+    'Removing old caniuse-lite from lock file\n' +
+    'Installing new caniuse-lite version\n' +
+    '$ npm install caniuse-lite\n' +
+    'Cleaning package.json dependencies from caniuse-lite\n' +
+    '$ npm uninstall caniuse-lite\n' +
+    'caniuse-lite has been successfully updated\n' +
+    '\n' +
+    'Target browser changes:\n' +
+    '- chrome 83\n' +
```
AleksandrSl, line 152: Vary flaky test, since we can't pin desired can i use version.
We can definitely check removed browsers versions.
For added one maybe some specific browserslist query can be used to minimize changes

test/update-db.test.js
```diff
@@ -133,3 +133,56 @@ if (!NODE_8 && (isInstalled('pnpm') || process.env.CI)) {
...
+    '+ chrome 86\n' +
+    '- edge 84\n' +
+    '+ edge 86\n' +
+    '- firefox 80\n' +
```
AleksandrSl, line 156: Maybe it's better to show difference for the same browser like:
```
- firefox 80, 79, 78
+ firefox 82, 81, 68
```
or 
```
- firefox 80 79 78
+ firefox 82 81 68
```

update-db.js
```diff
@@ -158,4 +200,32 @@ module.exports = function updateDB (print) {
   childProcess.execSync(del + ' caniuse-lite')
 
   print('caniuse-lite has been successfully updated\n')
+
+  var currentBrowsersList
+  if (!browsersListRetrievalError) {
+    try {
+      currentBrowsersList = getBrowsersList()
+    } catch (e) /* istanbul ignore next */ {
```
AleksandrSl, line 208: It's hard to cause error only on the second call to `getBrowsersList()` in tests, but i think such situation is possible.

update-db.js
```diff
@@ -158,4 +200,32 @@ module.exports = function updateDB (print) {
...
+      browsersListRetrievalError = e
+    }
+  }
+
+  if (browsersListRetrievalError) {
+    print(browsersListRetrievalError.stack)
```
AleksandrSl, line 214: I think error can be useful, e.g. we get error when npx is not installed, user can fix it based on error message

update-db.js
```diff
@@ -158,4 +200,32 @@ module.exports = function updateDB (print) {
...
+  } else {
+    var targetBrowserChanges = diffBrowsersLists(
+      oldBrowsersList,
+      currentBrowsersList
+    )
+    if (targetBrowserChanges) {
+      print('\nTarget browser changes:\n')
+      print(targetBrowserChanges + '\n')
+    } else {
+      print('\nNo target browser changes\n')
```
AleksandrSl, line 228: I think we can explicitly mention that there are no changes. Though user can deduce that there are no changes when there are no errors and `Target browser changes:` part is missing

update-db.js
```diff
@@ -65,6 +65,41 @@ function getLatestInfo (lock) {
...
+      result[entry[0]].push(entry[1])
+      return result
+    }, {})
+}
+
+function diffBrowsersLists (old, current) {
+  return Object.keys(old)
+    .sort()
+    .flatMap(function (browser) {
+      var intersection = old[browser].filter(function (version) {
```
AleksandrSl, line 86: I'm sure that more sophisticated algorithm can be used, but taking into account number of version to compare it will be overkill

test/update-db.test.js
AleksandrSl: I haven't implemented version sort, cause it can be complicated, e.g. to compare safari version like 12.0-12.1.
But if it's needed a simple one, e.g. based on the major version can be introduced

test/update-db.test.js
AleksandrSl: Problems started earlier than i thought.
node environment suddenly has fewer updates
Screenshot from 2020-11-08 00-30-08

ai: You can use `expect(…).toMatch(/…/)`

test/update-db.test.js
ai: Let’s keep as it and improve later

test/update-db.test.js
```diff
@@ -133,3 +133,56 @@ if (!NODE_8 && (isInstalled('pnpm') || process.env.CI)) {
...
+    '- firefox 79\n' +
+    '- firefox 78\n' +
+    '+ firefox 82\n' +
+    '+ firefox 81\n' +
+    '+ firefox 68\n' +
+    '- ios_saf 14.0\n' +
+    '- ios_saf 13.0-13.1\n' +
+    '- ios_saf 12.0-12.1\n' +
+    '+ ios_saf 14\n' +
+    '- opera 70\n' +
```
ai, line 166: I like that your group browsers by type 👍

update-db.js
```diff
@@ -65,6 +65,43 @@ function getLatestInfo (lock) {
   }
 }
 
+function getBrowsersList () {
+  return childProcess.execSync('npx browserslist').toString()
+    .trim()
+    .split('\n')
+    .map(function (line) { return line.trim().split(' ') })
```
ai, line 72:
```suggestion
    .map(function (line) {
      return line.trim().split(' ')
    })
```

update-db.js
```diff
@@ -65,6 +65,43 @@ function getLatestInfo (lock) {
...
+      var removedVersions = old[browser].filter(function (version) {
+        return intersection.indexOf(version) === -1
+      })
+      return removedVersions.map(function (version) {
+        return '- ' + browser + ' ' + version
+      }).concat(addedVersions.map(function (version) {
+        return '+ ' + browser + ' ' + version
+      }))
+    })
+    .reduce(function (result, array) { return result.concat(array) }, [])
```
ai, line 101:
```suggestion
    .reduce(function (result, array) {
      return result.concat(array)
    }, [])
```

update-db.js
ai: Note, that there are versions with multiple numbers `10-11` (see `ios` versions)

update-db.js
ai: I agree 👍

update-db.js
ai: Yeap. Good idea.

update-db.js
ai: Good idea. It shows that everything is safe.

test/update-db.test.js
AleksandrSl: As far as I understand node is an environment with he latest node version (15.1), and I still doesn't understand why it produces different results.
Running `npx browserslist` gives us the list using currently installed `caniuse-lite`, so the version of `caniuse-lite` is somehow different in different environments?

update-db.js
```diff
@@ -65,6 +65,43 @@ function getLatestInfo (lock) {
...
+      if (!result[entry[0]]) {
+        result[entry[0]] = []
+      }
+      result[entry[0]].push(entry[1])
+      return result
+    }, {})
+}
+
+function diffBrowsersLists (old, current) {
+  return Object.keys(old)
```
AleksandrSl, line 83: I also assume that the list of browsers itself won't change. e.g. if firefox was not in the old list it won't be in the current.
Though, my assumption can be wrong, since queries can be based on browsers usage statistics and they can change, so new browser can appear

test/update-db.test.js
AleksandrSl: I simplified the test to check presence of correctly formatted browser changes, since we can't tell what browser changes will pesent.

update-db.js
AleksandrSl: Fixed to remove this assumption

update-db.js
AleksandrSl: I would prefer the old solution, it's more readable and prefers pretty order. If we will find out that this places is really slow, then we can optimize it.

ai:
I marked the task as solved by you.

I will release this update tomorrow.

ai:
Released in 4.14.7

Issue #543, opened 2020-10-22
Respect --silent flag


tadjik1:
Hey there,
thanks a lot for the beautiful tool! Can you please improve the warning about outdated `caniuse-lite` version? At the moment it doesn't respect `--silent` flag and print messages in the console:
https://github.com/browserslist/browserslist/blob/4db22aed0f0453b73fabc87264bf7210d101075b/node.js#L373-L378

ai:
Why do you think this warning is not important for your case?

tadjik1:
I wouldn't say it's unimportant - but there should be a way to control which messages (or which levels) should be sent to the output. NPM has settings called `loglevel`, so `--silent` means that no messages should be printed. If you do not want to check NPM flags you can also provide your own environment variable or some other way to control this behavior.

ai:
What is default npm’s `loglevel`?

tadjik1:
it's `notice`: https://docs.npmjs.com/cli/v6/using-npm/config#loglevel

ai:
Yeap. Let’s respect `--silent`. Send PR.

extronics:
> Why do you think this warning is not important for your case?

I'm on a typescript/backend project using jest for tests. Since typescript + jest may use babel for transpilation (https://jestjs.io/docs/getting-started#using-typescript) which in turn uses `@babel/preset-env`, i'm now getting browserslist included in a backend-only project. So these warnings don't do anything for me (except maybe annoying me) since the code is not built for browsers.

I assume backend+typescript+jest it not that rare of a stack so there are probably a lot of users out there with this issue. 

It would be great if you could accept that your views on good practices may not hold for 100% of all cases where this library ends up in the dependencies. It's questionable behaviour already for a library somewhere deep down the dependency tree to start printing to the current process' standard error. Just imagine ANY non-human interpretation of process output suddenly receiving random outputs from libraries. So at least consider providing an escape hatch to silence this.

laurent22:
We work on a monorepo which includes a mobile and desktop app, so this caniuse-lite is not something we need but is probably included by a sub-package somewhere.

So we get a dozen of these useless warnings every time we build the Android app for example. And I tried to run the suggested command to finally get rid of them, but it's showing me an error about package.json, and I don't have all day to figure this out since we don't even need browserslist.

> Why do you think this warning is not important for your case?

In our case, indeed we know it's not important.

Aghassi:
[...]

PR #555, opened 2020-11-08
Respect --silent flag

ur300:
This PR addresses issue #543.
If `--silent` flag is there in the command line then `oldDataWarning` check omitted completely.


ai:
I reverted PR. The issue was not about Browserslist CLI.

It was about running `webpack --silent`.

So you need not change `cli.js`, but to change only `index.js` and check the global `process.argv`

PR #560, opened 2020-11-25
Throw error on missing environment

### LINKED ISSUE #559: browserslist returns default config for unrecognised environments
## Background

`browserslist` conveniently allows for keeping different lists, through different environments. These can be defined in any one of the locations that `browserslist` searches, such as `package.json`.

They can then be used by doing e.g.:

```js
browserslist(undefined, {env:'my-env'});
```

## Description of the issue

If the provided environment (`my-env`, in the example above) isn't defined everywhere, `browserslist` will return the default list of browsers, rather than providing any sort of indication that the list is missing. This can lead to hard-to-find bugs in client code if for some reason the `browserslist` configuration isn't present, such as not having been copied to a production environment.

## How to reproduce

- Run `browserslist(undefined, {env:'missing-env'});` without having `missing-env` defined anywhere
- Observe the response

### Expected result

An error or an empty list.

### Observed result

The default `browserslist` configuration. In my particular version:
```
[
  'and_chr 84',        'chrome 87',
  'chrome 86',         'chrome 85',
  'chrome 84',         'chrome 83',
  'edge 84',           'edge 83',
  'firefox 81',        'firefox 80',
  'firefox 79',        'firefox 78',
  'ios_saf 13.4-13.5', 'ios_saf 13.3',
  'opera 69',          'opera 68',
  'safari 13.1',       'safari 13'
]
```

## Next steps

I'm more than happy to work on a PR to fix this issue, but it would likely be best to reach a consensus first on what the correct behaviour should be:
- The current behaviour
- Throwing an error
- Returning an empty list

Thank you!

sgomes:
This PR introduces a check for missing environment configuration when explicitly providing an environment name, either by including it in `opts` or by providing it via the `BROWSERSLIST_ENV` environment variable.

It now throws an explicit error, whereas before it silently used the config defaults.

As discussed in #559, this is a breaking change.

Fixes #559.

### DIFF
--- a/node.js
+++ b/node.js
@@ -68,15 +68,20 @@ function pickEnv (config, opts) {
   if (typeof config !== 'object') return config
 
   var name
+
   if (typeof opts.env === 'string') {
     name = opts.env
   } else if (process.env.BROWSERSLIST_ENV) {
     name = process.env.BROWSERSLIST_ENV
-  } else if (process.env.NODE_ENV) {
-    name = process.env.NODE_ENV
-  } else {
-    name = 'production'
   }
+
+  if (name && name !== 'defaults' && !config[name]) {
+    throw new BrowserslistError(
+      'Missing config for Browserslist environment `' + name + '`.'
+    )
+  }
+
+  name = name || process.env.NODE_ENV || 'production'
 
   return config[name] || config.defaults
 }
--- a/test/main.test.js
+++ b/test/main.test.js
@@ -191,8 +191,12 @@ it('uses env options to browserlist config', () => {
   expect(browserslist(null, { path: CONFIG, env: 'development' }))
     .toEqual(['chrome 55', 'firefox 50'])
 
-  expect(browserslist(null, { path: CONFIG, env: 'test' }))
+  expect(browserslist(null, { path: CONFIG, env: 'defaults' }))
     .toEqual(['ie 11', 'ie 10'])
+
+  expect(() => {
+    browserslist(null, { path: CONFIG, env: 'test' })
+  }).toThrow(/Missing config for Browserslist environment/)
 })
 
 it('uses env options to package.json', () => {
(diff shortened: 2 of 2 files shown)


Semigradsky:
What if add an option about generate or not error in this case?
It can be not breaking change if this option will be disabled by default.
Tools which used `browserslist` can enable option in theirs major releases but `browserslist` will remain in `4.x` version.

sgomes:
Thanks for taking a look at this PR, @Semigradsky! I'm not sure I follow your suggestion, however. The idea is to only enable this in `browserslist 5.x`, since it's a breaking change. As such, any tools that use `browserslist 4.x` will keep the old behaviour, and only tools that get updated to `browserslist 5.x` will get the new behaviour. This is the correct use of a major version number, in my opinion.

Adding an option seems to me like a bad idea, because it could lead to different `browserslist 5.x`-enabled tools behaving differently for the same `browserslist` config and version.

Perhaps I'm misunderstanding your suggestion?

Semigradsky:
Don't worry, this is just my suggestion on how to release it in the near future)
Releasing a major version of `browserslist` can be painful.

ai:
@Semigradsky anyway we need major release for other features

hustla881 (approved):

Muradalsaaideh:
> This PR introduces a check for missing environment configuration when explicitly providing an environment name, either by including it in `opts` or by providing it via the `BROWSERSLIST_ENV` environment variable.
> 
> It now throws an explicit error, whereas before it silently used the config defaults.
> 
> As discussed in #559, this is a breaking change.
> 
> Fixes #559.

Muradalsaaideh:
> This PR introduces a check for missing environment configuration when explicitly providing an environment name, either by including it in `opts` or by providing it via the `BROWSERSLIST_ENV` environment variable.
> 
> It now throws an explicit error, whereas before it silently used the config defaults.
> 
> As discussed in #559, this is a breaking change.
> 
> Fixes #559. 

111e1f6ed230d7217e8aba7baa678e0f0dd52d14

Dasun2 (approved):

PR #562, opened 2020-11-26
Bail early if caniuse-lite is up to date

ialarmedalien:
This prevents the update-db script from going through the kerfuffle of updating `caniuse-lite` if the local version is already up to date.

### DIFF
--- a/test/update-db.test.js
+++ b/test/update-db.test.js
@@ -78,6 +78,23 @@ it('updates caniuse-lite for npm', async () => {
   expect(lock.dependencies['caniuse-lite'].version).toEqual(caniuse.version)
 })
 
+it('skips the update if caniuse-lite is up to date', async () => {
+  let dir = await chdir('update-npm', 'package.json', 'package-lock.json')
+  // update the dir to the current version
+  runUpdate()
+  let lock = JSON.parse(await readFile(join(dir, 'package-lock.json')))
+  expect(lock.dependencies['caniuse-lite'].version).toEqual(caniuse.version)
+
+  // rerun runUpdate; it should not try to install a new caniuse-lite version
+  expect(runUpdate()).toContain(
+    `Current version: ${ caniuse.version }\n` +
+    `New version:     ${ caniuse.version }\n` +
+    'caniuse-lite is up to date\n'
+  )
+  lock = JSON.parse(await readFile(join(dir, 'package-lock.json')))
+  expect(lock.dependencies['caniuse-lite'].version).toEqual(caniuse.version)
+})
+
 it('updates caniuse-lite without previous version', async () => {
   let dir = await chdir('update-missing', 'package.json', 'package-lock.json')
 
--- a/test/update-db.test.js
+++ b/test/update-db.test.js
@@ -6,8 +6,7 @@ let { nanoid } = require('nanoid/non-secure')
 let { tmpdir } = require('os')
 let { join } = require('path')
 
-let updateDd = require('../update-db')
-
+let updateDb = require('../update-db')
 const NODE_8 = process.version.startsWith('v8.')
 
 jest.setTimeout(10000)
@@ -33,7 +32,7 @@ async function chdir (fixture, ...files) {
 
 function runUpdate () {
   let out = ''
-  updateDd(str => {
+  updateDb(str => {
     out += str
   })
   return out
@@ -43,13 +42,53 @@ function isInstalled (cmd) {
   return execSync(`whereis ${ cmd }`).toString().trim() !== `${ cmd }:`
 }
 
+function checkRunUpdateContents (installedVersions, system) {
+  let addCmd = system + (system === 'yarn' ? ' add -W' : ' install')
+  let rmCmd = system + (system === 'yarn' ? ' remove -W' : ' uninstall')
+
+  expect(runUpdate()).toContain(
+    `Latest version:     ${ caniuse.version }\n` +
+    'Installed version' +
+    (installedVersions.indexOf(',') !== -1 ? 's:' : ': ') +
+    ` ${ installedVersions }\n` +
+    'Removing old caniuse-lite from lock file\n' +
+    'Installing new caniuse-lite version\n' +
+    `$ ${ addCmd } caniuse-lite\n` +
+    'Cleaning package.json dependencies from caniuse-lite\n' +
+    `$ ${ rmCmd } caniuse-lite\n` +
+    'caniuse-lite has been successfully updated\n'
+  )
+}
+
+function checkRunUpdateNoChanges () {
+  expect(runUpdate()).toContain(
+    `Latest version:     ${ caniuse.version }\n` +
+    `Installed version:  ${ caniuse.version }\n` +
+    'caniuse-lite is up to date\n'
+  )
+}
+
+const yarnLockfileVersions = 'caniuse-lite@^1.0.30000981, ' +
+  'caniuse-lite@^1.0.30001020, caniuse-lite@^1.0.30001030:'
+
+async function checkYarnLockfile (dir) {
+  let contents = (await readFile(join(dir, 'yarn.lock'))).toString()
+  expect(contents).toContain(
+    `${ yarnLockfileVersions }\n`
+  )
+  expect(contents).toContain(
+    `${ yarnLockfileVersions }\n` +
+    `  version "${ caniuse.version }"`
+  )
+}
+
 let caniuse = JSON.parse(execSync('npm show caniuse-lite --json').toString())
 
 it('throws on missing package.json', async () => {
   await chdir('update-missing')
   expect(runUpdate).toThrow(
     'Cannot find package.json. ' +
-    'Is it a right project to run npx browserslist --update-db?'
+    'Is this the right directory to run `npx browserslist --update-db` in?'
   )
 })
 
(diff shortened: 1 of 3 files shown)


update-db.js
```diff
@@ -180,17 +180,31 @@ module.exports = function updateDB (print) {
...
     print('Current version: ' + bold(red(current)) + '\n')
   }
   print(
-    'New version:     ' + bold(green(latest.version)) + '\n' +
+    'New version:     ' + bold(green(latest.version)) + '\n'
+  )
+
+  if (currentIsLatest) {
+    print(
+      'caniuse-lite is up to date\n'
```
ai, line 202: Can you show this line in green?

update-db.js
```diff
@@ -180,17 +180,31 @@ module.exports = function updateDB (print) {
...
   var oldBrowsersList
+  var currentIsLatest = false
   try {
     oldBrowsersList = getBrowsersList()
   } catch (e) {
     browsersListRetrievalError = e
   }
 
   if (typeof current === 'string') {
+    if (current === latest.version) {
```
ai, line 191: What is `yarn.lock` contains different versions of `caniuse-lite`?

ialarmedalien: Do you mean that you want a test for yarn, too?

ai: Nope.

Both `yarn.lock` and `package-lock.json` can contains a different versions of `caniuse-lite` (because of dublication problem of npm).

`--update-db` not only updates `caniuse-lite` version, but also remove dublicates.

So we need also to check that there is not different `caniuse-lite` versions in lock file.

ialarmedalien: It's an existing problem, then -- `getCurrentVersion` reports the first version of `caniuse-lite` that it comes across, rather than checking to see if multiple versions are installed in the various dependencies, and if so, what the version numbers are. There's also nothing in the test fixtures to ensure that `update-db.js` handles the duplicate situation correctly.

ai: > It's an existing problem, then -- getCurrentVersion reports the first version of caniuse-lite that it comes across

Yeap. But nod duplication will be fixed anyway (even if the bigger version from the lock file is the latest). This bug affects only CLU output.

After this PR `--update-db` will not fix `caniuse-lite` duplicates if the bigger version from the lock file is the latest.

ialarmedalien: How about if there are multiple versions of `caniuse-lite` because `package.json` specifies a particular version? The current search and replace approach takes no account of this (edge) case.

ai: In the current approach, we remove all `caniuse-lite` mentions from the lock and force lock file update. `npm` or `yarn` will download the latest `caniuse-lite` for dependencies requirements.

So if the app locked the version, it will have the locked version. But other tools will have the latest since they didn’t lock the version. It is an expected approach since `package.json` is not a way to lock dependency for the whole dependencies tree.

ai: I suggest improving the check. We should find the smallest `caniuse-lite` versions in the lock file and run the update script if the smallest is not equal to the latest.

test/update-db.test.js
```diff
@@ -60,126 +97,105 @@ it('throws on missing lockfile', async () => {
   )
 })
 
-it('updates caniuse-lite for npm', async () => {
-  let dir = await chdir('update-npm', 'package.json', 'package-lock.json')
+it('shows target browser changes', async () => {
```
ialarmedalien, line 100: moved this up - it is unchanged apart from being in a different position in the file

test/update-db.test.js
```diff
@@ -60,126 +97,105 @@ it('throws on missing lockfile', async () => {
...
+  expect(runUpdate()).toMatch(
+    // eslint-disable-next-line max-len
+    /(Target browser changes:\n([+-] \w+ [\d.-]+\n)+)|(No target browser changes)/
   )
 
   let lock = JSON.parse(await readFile(join(dir, 'package-lock.json')))
   expect(lock.dependencies['caniuse-lite'].version).toEqual(caniuse.version)
 })
 
+it('shows error when browsers list can\'t be retrieved', async () => {
```
ialarmedalien, line 113: Same with this test - moved up the file

update-db.js
```diff
@@ -41,24 +42,6 @@ function detectLockfile () {
   }
 }
 
-function getCurrentVersion (lock) {
```
ialarmedalien, line 44: This is now done as part of updating the lock file.

update-db.js
```diff
@@ -117,66 +100,100 @@ function diffBrowsersLists (old, current) {
...
+  return metadata
 }
 
-function deletePackage (node) {
+function deletePackage (node, metadata) {
   if (node.dependencies) {
-    delete node.dependencies['caniuse-lite']
+    if (node.dependencies['caniuse-lite']) {
+      var version = node.dependencies['caniuse-lite'].version
+      metadata.versions[version] = true
```
ialarmedalien, line 114: store the versions of `caniuse-lite` found in the file

update-db.js
```diff
@@ -117,66 +100,100 @@ function diffBrowsersLists (old, current) {
...
 }
 
-module.exports = function updateDB (print) {
-  var lock = detectLockfile()
+var yarnVersionRe = new RegExp('version "(.*?)"')
+
+function updateYarnLockfile (lock, latest) {
+  var blocks = lock.content.split(/(\n{2,})/).map(function (block) {
+    return block.split('\n')
+  })
```
ialarmedalien, line 129: It's easier to parse the yarn lock file if you split it up into blocks separated by two or more `\n`s, which are then split into lines. Apart from the first block (header material), the blocks have the same format.

update-db.js
```diff
@@ -117,66 +100,100 @@ function diffBrowsersLists (old, current) {
...
+        )
+        lines[3] = lines[3].replace(
+          /integrity .+/, 'integrity ' + latest.dist.integrity
+        )
+      }
+    }
+  })
+  var content = blocks.map(function (lines) {
+    return lines.join('\n')
+  }).join('')
```
ialarmedalien, line 150: Restore original form

update-db.js
```diff
@@ -117,66 +100,100 @@ function diffBrowsersLists (old, current) {
...
+  return { content: content, versions: versions }
+}
+
+function updatePnpmLockfile (lock, latest) {
+  var versions = {}
+  var caniusePackage
+  var content = yaml.safeLoad(lock.content)
```
ialarmedalien, line 157: Make life easier by using a yaml parser.

update-db.js
```diff
@@ -186,17 +203,35 @@ module.exports = function updateDB (print) {
...
   }
 
-  if (typeof current === 'string') {
-    print('Current version: ' + bold(red(current)) + '\n')
+  print('Checking whether caniuse-lite needs to be updated...\n')
+  print(
+    'Latest version:       ' + bold(green(latest.version)) + '\n'
+  )
+
+  var lockfileData = updateLockfile(lock, latest)
```
ialarmedalien, line 211: This is in the form 
```
{ content: <string of lock file content>, versions: { <versionNo>: true, <versionNo2>: true, ... } }
```

update-db.js
ialarmedalien: I've rewritten `update-db.js`to find all versions of `caniuse-lite` in the lock file, and rewrite the file and trigger an update if there are any versions other than the latest present.

package.json
```diff
@@ -22,6 +22,7 @@
     "colorette": "^1.2.1",
     "electron-to-chromium": "^1.3.591",
     "escalade": "^3.1.1",
+    "js-yaml": "^3.14.0",
```
ai, line 25: It will increase `browserslist` dependencies by 3 dependencies (`js-yaml`, `argparse`, `esprima`). It is too much for `--update-db` which you need only one per month.

Can we do it without library?

test/update-db.test.js
```diff
@@ -60,126 +97,105 @@ it('throws on missing lockfile', async () => {
...
+  checkRunUpdateContents('1.0.30001030', 'npm')
 
   let lock = JSON.parse(await readFile(join(dir, 'package-lock.json')))
-  expect(lock.dependencies['caniuse-lite']).toBeUndefined()
+  expect(lock.dependencies['caniuse-lite'].version).toEqual(caniuse.version)
+})
+
+it('skips the npm update if caniuse-lite is up to date', async () => {
+  let dir = await chdir('update-npm', 'package.json', 'package-lock.json')
+  // update the dir to the current version
```
ai, line 144: Can you remove comments? I prefer to change function names if the code is not readable without comments.

update-db.js
```diff
@@ -117,66 +100,100 @@ function diffBrowsersLists (old, current) {
...
+  if (lock.mode === 'npm') {
+    return updateNpmLockfile(lock, latest)
+  } else if (lock.mode === 'yarn') {
+    return updateYarnLockfile(lock, latest)
+  }
+  return updatePnpmLockfile(lock, latest)
+}
+
+module.exports = function updateDB (print) {
+  print('running updateDb!\n')
```
ai, line 195: Is the next command slow? If so, it will be better to explain what happening like `Reading lock file…`

update-db.js
```diff
@@ -186,17 +203,35 @@ module.exports = function updateDB (print) {
...
+    )
+    return
+  }
+
+  if (caniuseVersions.length === 0) {
+    caniuseVersions[0] = 'none'
   }
   print(
-    'New version:     ' + bold(green(latest.version)) + '\n' +
+    'Installed version(s): ' +
```
ai, line 226: Let’s print `version` or `versions` according to `caniuseVersions === 1`

update-db.js
```diff
@@ -186,17 +203,35 @@ module.exports = function updateDB (print) {
     browsersListRetrievalError = e
   }
 
-  if (typeof current === 'string') {
-    print('Current version: ' + bold(red(current)) + '\n')
+  print('Checking whether caniuse-lite needs to be updated...\n')
```
ai, line 206: We do not need to print step description if it is very fast

ai:
@ialarmedalien you are doing a lot of work. Thanks and sorry for my perfectionism (perfectionism is important for so popular project).

update-db.js
ialarmedalien: Removed.

package.json
ialarmedalien: It's much easier to parse the pnpm lock file using a yaml parser, rather than with regular expressions. I changed the way that the pnpm lock file is updated to account for cases where more than one version of `caniuse-lite` is installed (delete nodes and let pnpm rewrite the lock file), and that would be a lot messier to do via naïve line-by-line string parsing. In the grand scheme of things, I think having code that is easier to read, update, and maintain trumps a small dependency list.

There are other yaml parsers that could be used:
* [yaml](https://www.npmjs.com/package/yaml) -- 0 deps!
* [yamljs](https://www.npmjs.com/package/yamljs) -- 2 deps

ai: `pnpm` use case is so rare. We can’t add one dependency just to simplify work for 1% of users. 99% of users will download an unnecessary package.

I think users' comfort is much more important than the developer’s comfort. Let’s use RegExp for now.

In the next major release of Browserslist, I will move `update-db` to a separated package and will clean up the code with Yaml parsers there.

ialarmedalien: How many users even look at package dependencies, though? I certainly didn't when I started using browserslist -- I just followed the instructions at the postcss website which linked to the best practices section of the readme. If you are worried about having an extra install for some users, the yaml parser could be a peer or optional dependency that gets installed if it isn't present during the pnpm workflow.

ai: > How many users even look at package dependencies, though? I certainly didn't when I started using browserslist

The problem is big `node_modules` containing a lot of unnecessary code.

Keeping optional dependency will force us to show a good error message on the missed YAML parser. It is still extra work. It will be better to use regexp.

update-db.js
```diff
@@ -117,66 +96,106 @@ function diffBrowsersLists (old, current) {
...
+}
+
+function updatePnpmLockfile (lock, latest) {
+  var versions = {}
+  var lines = lock.content.split('\n')
+  var i
+  var j
+  var lineParts
+
+  for (i = 0; i < lines.length; i++) {
```
ialarmedalien, line 157: parse pnpm yaml by hand

package.json
ialarmedalien: I disagree. Hand-parsing is always more fragile.

I have rewritten the pnpm code and removed the yaml parser dependency.

ialarmedalien:
@ai is there more that you want done to this PR?

ai:
Thanks for noticing, totally forgot about this PR (too many open source projects)

Released in 4.16.2.

PR #588, opened 2021-03-19
Remove a erroneous white space character

koddsson:
Title pretty much says it all 😄


ai:
It was added for a reason because prefix contain a space 🙂

PR #591, opened 2021-04-07
Update DB - Support for artifactory (No integrity)

### LINKED ISSUE #592: Update DB not working with Artifactory NPM mirror
This is because Artifactory does not expose an integrity field like NPM does.

I tried to do a very simple fix which works for me locally, but for some reason it fails on Node 10 (Though fine on Node 8 and 15, odd)

https://github.com/browserslist/browserslist/pull/591

ittaibaratz:
Fixes #592 

This makes the update db tool work when using artifactory to mirror the NPM registry.

In this case, integrity is not in latest.dist, shasum is used instead.

Leaving integrity out in this case makes yarn add it automatically.

### DIFF
--- a/update-db.js
+++ b/update-db.js
@@ -135,9 +135,9 @@ function updateYarnLockfile (lock, latest) {
         lines[2] = lines[2].replace(
           /resolved "[^"]+"/, 'resolved "' + latest.dist.tarball + '"'
         )
-        lines[3] = lines[3].replace(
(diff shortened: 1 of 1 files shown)


ai:
Yeap. CI was broken because of the `npm` update on GitHub Actions. I fixed it.

Thanks for the fix. I will release it with another fix a few hours later.

ai:
The fix was released in 4.16.4

PR #603, opened 2021-06-01
Update README.md

eduplessis:
Confusing because it make it like the `Firefox > 20` was under the IOS parameter


ai:
I grouped queries by type. Maybe I should add titles to groups to make grouping system more clear.

ai:
Groups titles ad3befe

PR #630, opened 2021-09-28
Bumped caniuse-lite version to include Safari 15 release

leereichardt:
This PR bumped the version of the caniuse-lite package as this has been updated to bring Safari 15 out of technical preview with a full release as showing on caniuse.com


ai:
You do need to bump this version, since it already has `^` in version. `browserslist` will use the latest `caniuse-lite`.

PR #638, opened 2021-11-02
fix: check file path when resolving shared configs

adalinesimonian:
Fixes stylelint/vscode-stylelint#108

Currently, only the current working directory is checked when resolving shareable configs. However, in certain instances, browserslist may be running outside of the directory in which the file referencing the config is found. This fix adds the context's path as a fallback when resolving shareable configs.

### DIFF
--- a/node.js
+++ b/node.js
@@ -161,7 +161,7 @@ module.exports = {
       checkExtend(name)
     }
     // eslint-disable-next-line security/detect-non-literal-require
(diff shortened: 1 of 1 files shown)


ai:
LGTM. I am going to sleep right now. Tomorrow I will review it again and very likely release a new patch release.

ai:
Thanks. Released in 4.17.6 (sorry for delay, it took a while to fix CI because of new npm lock file format)

PR #647, opened 2021-11-15
Fix edge case of `since` query (fix #646)

### LINKED ISSUE #646: Unreleased versions will be included when querying `since 1970`
When querying `since 1970`, it will be parsed into the UNIX timestamp `0`. On the other hand, in `caniuse-lite` data, release date of unreleased versions are `null`.

Now, when processing `since` query, there is a comparison logic: https://github.com/browserslist/browserslist/blob/29cecf978e78d4e3daab99809e8816488fcf8fb7/index.js#L212 .

In JavaScript, `null >= 0` will be evaluated to `true`, which causes unreleased versions will be included in the result when querying `since 1970`.

I'm not sure whether it's a bug or not. I found this problem when I was implementing `since` query in [browserslist-rs](https://github.com/g-plane/browserslist-rs), and currently I implemented this feature/bug as-is: https://github.com/g-plane/browserslist-rs/blob/35ed72f516ab1230d0c102169a9bbe25f493fedb/src/queries/since.rs#L53-L62 .

g-plane:
This PR adds a check that makes sure release date won't be `null`.

### DIFF
--- a/index.js
+++ b/index.js
@@ -209,7 +209,8 @@ function filterByYear(since, context) {
     var data = byName(name, context)
     if (!data) return selected
     var versions = Object.keys(data.releaseDate).filter(function (v) {
-      return data.releaseDate[v] >= since
+      var date = data.releaseDate[v]
+      return date !== null && date >= since
     })
     return selected.concat(versions.map(nameMapper(data.name)))
   }, [])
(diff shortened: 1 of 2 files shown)


ai:
Thanks. Released in 4.18.1.

PR #648, opened 2021-11-15
Make `cover` query case-insensitive

g-plane:
This PR makes `cover` query case-insensitive, which allows:

```
Cover 1%
Cover 1% In My Stats
Cover 1% in us
Cover 1% In Alt-US
```


ai:
Thanks. Released in 4.18.1.

PR #664, opened 2021-12-13
feat: added option to throw on missing env

fivethreeo:
Based on https://github.com/browserslist/browserslist/pull/560

* Added `throwOnMissing` option to browserslist
* Added tests for `throwOnMissing` option to browserslist
* Added jsdoc for option but found no script for updating .d.ts
* Coverage not full but can't figure out why

### DIFF
--- a/node.js
+++ b/node.js
@@ -82,6 +82,14 @@ function pickEnv(config, opts) {
     name = 'production'
   }
 
+  if (opts.throwOnMissing) {
+    if (name && name !== 'defaults' && !config[name]) {
+      throw new BrowserslistError(
+        'Missing config for Browserslist environment `' + name + '`.'
+      )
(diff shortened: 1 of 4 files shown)


node.js
```diff
@@ -82,6 +82,14 @@ function pickEnv(config, opts) {
     name = 'production'
   }
 
+  if (opts.throwOnMissing) {
```
ai, line 85: Where do we set default value?

ai:
Thanks. Released in 4.19.

fivethreeo:
Is the d.ts built from jsdoc? Seems to be missing the option in the typing.

ai:
Nope. We have separated `.d.ts` file. Can you fix it in PR?

PR #671, opened 2022-03-05
Add `last N node versions` (with major) support

### LINKED ISSUE #410: bug or expected: `last 1 Node version` throws `browserslist: Unknown browser Node`
As title, I think it should make sense like `last 1 Chrome version`.

g-plane:
This PR added support:

```
last 2 node versions
```

and 

```
last 2 node major versions
```

Close #410 .

### DIFF
--- a/test/node.test.js
+++ b/test/node.test.js
@@ -345,10 +345,36 @@ test('supports range selection', () => {
 
 test('supports last versions for Node.js', () => {
   is(browserslist('last 2 node versions').length >= 1, true)
+
+  browserslist.nodeVersions = ['16.0.0', '16.1.0', '17.0.0', '17.1.0']
+  equal(browserslist('last 3 node versions'), [
+    'node 17.1.0',
+    'node 17.0.0',
+    'node 16.1.0'
+  ])
 })
 
 test('supports last major versions for Node.js', () => {
   is(browserslist('last 2 node major versions').length >= 1, true)
+
+  browserslist.nodeVersions = [
+    '14.0.0',
+    '14.1.0',
+    '15.0.0',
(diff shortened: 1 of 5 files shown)


test/node.test.js
```diff
@@ -343,4 +343,12 @@ test('supports range selection', () => {
   )
 })
 
+test('supports last versions for Node.js', () => {
+  is(browserslist('last 2 node versions').length >= 1, true)
+})
+
+test('supports last major versions for Node.js', () => {
+  is(browserslist('last 2 node major versions').length >= 1, true)
```
ai, line 351: This test doesn’t test what Browserslist return. What if it return `node 4` or invalid `node2`?

We need to mock data.,\

g-plane: That `node.test.js` isn't using mock data. Should we create mock data for all tests in that file or just these two new tests?

ai: I think it will be better to have 2 types of test:

1. With real data (in case if npm package will break API)
2. With mock data

So let’s keep this `.length` test and add an 2 extra tests with mock data.

g-plane: How can we mock data?

ai: Have no idea. What do you suggest?

g-plane: I also don't know how to do, since uvu doesn't support mocking.

ai: You can use some another tool for mocking or change Browserslist API to add mocking feature like we do with browser data

ai:
And the last thing can you add new query to `README.md`?

ai:
Thanks. Released in 4.20.

PR #672, opened 2022-03-05
feat: linter

dangreen:
WIP


index.d.ts
```diff
@@ -142,6 +142,18 @@ declare namespace browserslist {
...
+   * Lint browserslist config.
+   *
+   * @param [queries=browserslist.defaults] Browser queries.
+   * @param [opts] Browserslist options.
+   * @returns Lint report.
+   */
+  function lint(
+    queries?: string | readonly string[] | null,
+    opts?: browserslist.Options
+  ): string
```
ai, line 155: I think we need to return `string[]`. We may have multiple errors. We need a simple way to check that there is no error.

parser.js
```diff
@@ -0,0 +1,91 @@
+var QUERY_OR = 1
+var QUERY_AND = 2
+
+var QUERY_REGEXP = {
```
ai, line 4: Let’s use plural here

index.d.ts
dangreen: @ai done

parser.js
dangreen: @ai done

ai (approved):

ai:
LGTM. I will try to merge and move it to `browserslist-tool` when I finish my current Logux Server task.

ai:
Sorry for the delay. I haven’t enough energy.

PR #709, opened 2022-07-04
Add "since COVID" and "post-COVID" queries

cutiful:
The cultural importance of COVID cannot be overstated. It marks the beginning of a new era, one in which an action as simple as going outside might be deadly. Because of the major changes it introduced into our lives, as well as trauma that cannot be forgotten, many use it as a date signifier even in contexts unrelated to the pandemic itself. Consider the phrases ["pre-COVID"](https://trends.google.com/trends/explore?date=today%205-y&geo=US&q=pre-covid) and "post-COVID".

I think it is appropriate to add "since COVID" and "post-COVID" as valid Browserslist queries. I'll understand if you disagree, though.

### DIFF
--- a/index.js
+++ b/index.js
@@ -10,6 +10,7 @@ var env = require('./node') // Will load browser.js in webpack
 
 var YEAR = 365.259641 * 24 * 60 * 60 * 1000
 var ANDROID_EVERGREEN_FIRST = 37
+var COVID_QUERY = "since 2019-12-01"; // First known case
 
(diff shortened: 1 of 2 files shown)


ai:
I like the idea, but when I think about supporting this code, I do not want to spend time on this.

Also, it doesn’t fit the hedgehog style of Browserslist to do it just for fun.

cutiful:
Understandable. Have a good day :)

PR #717, opened 2022-08-16
Update browsersl.ist site URL in README

a-chabin:
I have updated the link to the new site [browsersl.ist](https://browsersl.ist) in README.md. The old address [browserl.ist](https://browserl.ist) is not available


ai:
Hm http://browserl.ist/ works but not https://browserl.ist/

I will look how to fix redirect

PR #728, opened 2022-09-08
include BROWSERSLIST_IGNORE_OLD_DATA in the README

ahmadnassri:
the env variable info was missing from the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -671,6 +671,12 @@ with [environment variables]:
   BROWSERSLIST_DANGEROUS_EXTEND=1 npx webpack
    ```
 
+* `BROWSERSLIST_IGNORE_OLD_DATA` to disable updating the Browserlist DB
+
+   ```sh
+  BROWSERSLIST_IGNORE_OLD_DATA=1 npx webpack
+   ```
+
 [environment variables]: https://en.wikipedia.org/wiki/Environment_variable
 
 


ahmadnassri:
I understand there is a value in keeping the db updated, but I'm currently faced with an issue where i'm getting this error in a CLI project, because of nested dependency chain includes `browserslist` ... and it's failing tests because of it (since those tests are parsing CLI output)...

my scenario might be a special condition, but browserslist is included in the dependency tree of many tools, and this console message will be irrelevant to many projects, suppressing should be an acceptable Behaviour.

ai:
Can you describe your use case?

What is this tool? How do you use it for your project?

ahmadnassri:
image

`tap` and `nyc` are the most common use-case I come across, my projects (not shown, preliterary) are node libraries, or cli tools ... `browserlist` is not even relevant to my projects, but when running tests, in this case with `tap` and `nyc` I hit this db upgrade issue

ai:
The problem is that Browserslist will show this warning only when it was called.

It means Babel really used to add polyfills to your tests. It means that with `caniuse-lite` update you can speed up your tests with a small fraction.

ahmadnassri:
sure, the problem is `nyc` and `tap` are not my projects, they are tools I use, and they in-turn use babel indirectly ...

ai:
Maybe the solution is to send PR to `nyc` or `tap` and avoid calling Browserslist? If you don’t need polyfills, this call just use resources.

ahmadnassri:
sure, I'll got and PR everybody who uses bable/core ... vs. just exposing the environment variable that's already implemented in the documentation ...

ai:
Adding this env to docs will do more harm.

1% who really need it vs 99% who will use it just to hide warning instead of making website faster.

PR #729, opened 2022-09-12
Fix links of `postcss-preset-env` and `postcss-normalize`

y-yagi:
Both links are specified to old repositories.  This fixed links to the correct repositories.


ai:
Thanks

PR #739, opened 2022-11-16
feat: make recommended defaults easier to copy

dpschen:
By including the whole recommended string we prevent wrong usage.

### DIFF
--- a/README.md
+++ b/README.md
@@ -161,13 +161,12 @@ These extensions will add syntax highlighting for `.browserslistrc` files.
     ]
   ```
 
-* If you want to change the default set of browsers, we recommend combining
-  `last 2 versions`, `not dead` with a usage number like `> 0.2%`. This is
-  because `last n versions` on its own does not add popular old versions, while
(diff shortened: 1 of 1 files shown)


ai:
LGTM

README.md
```diff
@@ -161,10 +161,10 @@ These extensions will add syntax highlighting for `.browserslistrc` files.
     ]
   ```
 
-* If you want to change the default set of browsers, we recommend combining
-  `last 2 versions`, `not dead` with a usage number like `> 0.2%`. This is
+* If you want to change the default set of browsers, we recommend including
+  `last 2 versions and not dead > 0.2%`. This is
```
dpschen, line 165: I was actually unsure from the description if the intended combinator ist `and` or `or`

ai: `X, not dead` == `X and not dead`

I recommend using `last 2 versions, > 0.2%, not dead`

ai:
Thanks

PR #763, opened 2023-05-27
Fix Android time queries with mobileToDesktop

### LINKED ISSUE #762: Time queries with mobileToDesktop do not work for Android
I would expect the following queries all to produce the same results:

- `last 1000 versions`
- `last 100 years`
- `since 1969`

Instead the latter 2 are different with `mobileToDesktop = true` because Android versions are not back-filled to the first evergreen version as with the first query.

steverep:
Fixes #762.  The release dates were not being cloned along with the version data.

### DIFF
--- a/index.js
+++ b/index.js
@@ -270,6 +270,11 @@ function normalizeAndroidVersions(androidVersions, chromeVersions) {
 function normalizeAndroidData(android, chrome) {
   android.released = normalizeAndroidVersions(android.released, chrome.released)
   android.versions = normalizeAndroidVersions(android.versions, chrome.versions)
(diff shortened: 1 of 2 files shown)


ai:
Can you also add tests to avoid regressions in the future?

steverep:
Done.  I left a placeholder for an extra test that should pass but doesn't because `mobileToDesktop` also doesn't work with usage queries.

ai:
LGTM for now. But it is too late and I am planning to release it tomorrow to be able to review it in the morning.

ai:
Thanks. Released in 4.21.6.

PR #765, opened 2023-05-28
Fix computation of last queries for Android

### LINKED ISSUE #764: Android "last" queries incorrectly include Android 36
Found another issue while fixing #762.  The code correctly tries to limit Android to a minimum of 37, but the computations are flawed.  It incorrectly assumes that the version history is always just incremented by 1, but this was broken when Chrome/Android 82 was abandoned during the pandemic.  The result is that a minimum of 36 is being used instead.

Fix pending...

steverep:
Fixes #764 by directly finding the correct array index.  This simply assumes that Chrome 37 exists and that the latest release of Chrome equals that of Android, rather than assuming anything about the version history.

A few tests failed after the change, but they are all flawed.  After adjusting, I think there is good coverage.

Note Prettier made a few extra changes so you might want to turn off white space on the diffs.

### DIFF
--- a/index.js
+++ b/index.js
@@ -9,7 +9,7 @@ var parse = require('./parse')
 var env = require('./node') // Will load browser.js in webpack
 
 var YEAR = 365.259641 * 24 * 60 * 60 * 1000
-var ANDROID_EVERGREEN_FIRST = 37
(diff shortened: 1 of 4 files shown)


ai:
Thanks. Released in 4.21.7.

PR #767, opened 2023-06-12
Fix feature support queries with mobileToDesktop

### LINKED ISSUE #761: mobile-to-desktop option is negated by feature support
The `mobile-to-desktop` flag appears to be completely negated by adding an intersection with a feature query.  For example:

```
$ npx browserslist --mobile-to-desktop "defaults"
and_chr 113
and_chr 112
and_ff 113
and_ff 112
and_qq 13.1
and_uc 13.4
android 113
android 112
...
```

But adding any `supports` feature query produces different results:

```
$ npx browserslist --mobile-to-desktop "defaults and supports es5"
and_chr 113
and_ff 113
and_qq 13.1
and_uc 13.4
android 113
...
```

I get why this is happening since `caniuse` doesn't keep data on older mobile versions, but it seems to me that the entire point of the `mobileToDesktop` option is to assume that mobile version numbers act like their desktop counterparts, so this is definitely unexpected and should be fixed IMO.

steverep:
Fixes #761 by adding support for `mobileToDesktop` to feature queries.

- Changed the feature cache to a 2 dimensional object for each browser and version to make the implementation much simpler
- Uses existing function to get the mobile to desktop data
- Only looks to the desktop version's support data when (a) there is no mobile data and (b) the feature is supported by the latest mobile version
- Added a couple specific mobile to desktop tests and made the others more specific for the common data

Note the way the existing tests were messing with the `cache` property doesn't really work.  This is evident if you just try to use the same feature from one test to the next (the first cache setting is always used).  I didn't bother to try to make it work because I think it would involve exporting the cache itself, so I just removed some of that and use a different feature for each test.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -124,7 +124,9 @@ declare namespace browserslist {
 
   let cache: {
     [feature: string]: {
-      [name: string]: 'y' | 'n'
+      [name: string]: {
+        [version: string]: string
+      }
     }
   }
 
--- a/index.js
+++ b/index.js
@@ -302,6 +302,13 @@ function filterAndroid(list, versions, context) {
   return list.slice(nEvergreen - 1 - versions)
 }
 
+function isSupported(flags) {
+  return (
+    typeof flags === 'string' &&
+    (flags.indexOf('y') >= 0 || flags.indexOf('a') >= 0)
+  )
+}
+
 function resolve(queries, context) {
   return parse(QUERIES, queries).reduce(function (result, node, index) {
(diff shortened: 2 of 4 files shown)


Semigradsky:
Just a reminder that need to change typings also:
https://github.com/browserslist/browserslist/blob/312d763173f9080d73cbef7c56966ce600ab9f4a/index.d.ts#L125-L129

ai:
Released at 4.21.8.

PR #768, opened 2023-06-14
Fix invalid results for Opera Mobile

### LINKED ISSUE #766: Opera mobile is invalid with mobileToDesktop
Unlike Android, Chrome, and Firefox, Opera's mobile version history does not mirror that of Opera desktop.  With the `mobileToDesktop` option enabled, the results are invalid:

```
$ ./cli.js --mobile-to-desktop  "last 2 op_mob versions"
op_mob 98
op_mob 97
```

Those versions don't exist - the latest is 75.  Furthermore, time queries or feature queries (after fixing #761) for this latest version will then use data for Opera desktop 75, which is much older and based on Chromium 89 versus 112.

I would propose removing it from the mobile to desktop mapping unless the versions can be properly mapped in terms of release date and/or chromium version.

A filter similar to the one used for Android is also needed to stop "last_version" queries from grabbing very old versions without `mobileToDesktop`:

```
$ ./cli.js  "last 2 op_mob versions"
op_mob 73
op_mob 12.1
```

steverep:
Fixes #766 by:

- Remove use of Opera desktop when using `mobileToDesktop: true` as the release history is quite different
- Adds logic similar to Android to handle the CanIUse version jump from 12.1 to latest
- Removes unused `mapVersions` function
- Adjusts existing tests and adds one to test handling the version jump for "last" queries
- Removes aliasing version 59 to 58 (both of these versions exist but cannot be selected anymore anyway)

Note that neither the android nor op_mob jumps are handled perfectly when querying only "major" versions, but that's truly an edge case.  I can follow up later with a fix for that.

### DIFF
--- a/index.js
+++ b/index.js
@@ -10,6 +10,7 @@ var env = require('./node') // Will load browser.js in webpack
 
 var YEAR = 365.259641 * 24 * 60 * 60 * 1000
 var ANDROID_EVERGREEN_FIRST = '37'
+var OP_MOB_BLINK_FIRST = 14
 
 // Helpers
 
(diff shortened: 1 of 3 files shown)


ai:
Looks very nice.

Do you need to release it now?

steverep:
No rush to release if you're waiting on something else.  My end goal is getting a regex for serving modern/legacy builds that is not overly restrictive on mobile.  I think I'm good with this repo for now and need to work on https://github.com/browserslist/browserslist-useragent-regexp/issues/1484.

ai:
Released in 4.21.9.

PR #770, opened 2023-06-25
use exact version to have reproducible builds

ValentinBossi:
when checking out an old browserslist version - e.g from 2020 - and running browserslist("last 2 chrome versions") it prints now the browser versions from 2020 not from 2023.


ai:
It will force me to release a version of Browserslist on every Can I Use release (about every day). I don't have so much resources.

ai:
You should use npm lock file to have reproductible results

PR #787, opened 2023-09-03
Add `fully supports` and `partially supports` queries

BPScott:
### Context 

Caniuse features can be marked as "fully supported", "partially supported" and a few other values that aren't relevant to this reminder/primer (see the [stats key in caniuse docs](https://github.com/Fyrd/caniuse/blob/main/CONTRIBUTING.md#supported-changes)). For instance for the [`es6`](https://caniuse.com/es6) feature, partial support began with Chrome 21 but wasn't marked as fully supported till Chrome 51.

Browserslist's `supports` query includes browsers with both full and partial support for a given feature. Thus `supports es6` includes Chrome 21 and up. This is somewhat surprising to me as if you'd try to use es6 features in Chrome 21 things would go very wrong very quickly. I hope to think of browserslist's supports query as "this is a baseline from which you can assume full support for a feature, so stuff like babel won't need to transpile/polyfill this at all", but thanks to it's partial support that's not quite what it does.

I'd like a way to make a supports query that only considers full support, instead of partial and full support. I appreciate that changing the behaviour of supports after so long would be an unexpected change to the point of it being a breaking change and so a new query would be needed.

This PR is an approach to solve that problem, but I'd love some feedback on naming.

### This PR

This PR adds a `fully supports` query, that acts like `supports` except it only considers full browser support. As a result `fully supports es6`'s first Chrome version is Chrome 51. 

I'm not super tied to the `fully supports` name and it feels a bit clunky. In an ideal world it'd feel nicer for `supports` to be for full support, and then we have a modifier if we want to include partial support, but I know that's almost certainly too disruptive to ship. I would welcome suggestions for any syntax that you'd prefer.

One alternative is having `supports es6 fully` - where fully is a modifier of the existing supports query, kinda like how you can do `>5%` and `>5% in US`.

Another alternative is betting on this prefix more, and adding `partially supports` as a query too (with the same behaviour as `supports` currently and pushing people to use `fully supports` and `partially supports` instead of `supports`. This kinda leans into the magical future world where we could change the default behaviour of `supports` to be "full support", and then people have to opt in if they want to include partial support.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ when you add the following to `package.json`:
 
 ```json
   "browserslist": [
-    "defaults and partially supports es6-module",
+    "defaults and fully supports es6-module",
     "maintained node versions"
   ]
 ```
@@ -29,7 +29,7 @@ Or in `.browserslistrc` config:
 ```yaml
 # Browsers that we support
 
-defaults and partially supports es6-module
+defaults and fully supports es6-module
 maintained node versions
 ```
 
@@ -273,18 +273,20 @@ You can specify the browser and Node.js versions by queries (case insensitive):
     to PhantomJS runtime.
 * `extends browserslist-config-mycompany`: take queries from
   `browserslist-config-mycompany` npm package.
-* `fully supports es6-module`: browsers with full support for specific features.
-  `es6-module` here is the `feat` parameter at the URL of the [Can I Use]
-  page. For example `fully supports css-grid` will omit Edge 12-15, as those browser versions are marked as [having partial support].
-  A list of all available features can be found at [`caniuse-lite/data/features`].
-* `partially supports es6-module`: browsers with full or partial support for specific features.
-  `es6-module` here is the `feat` parameter at the URL of the [Can I Use]
-  page. For example `fully supports css-grid` will include Edge 12-15 support, as those browser versions are marked as [having partial support].
-  A list of all available features can be found at [`caniuse-lite/data/features`].
-* `supports es6-module`: browsers with full or partial support for specific features. An alias for `partially supports es6-module`.
+* By browser support:<br>
+  In these example queries `es6` and `es6-module` are the the `feat` parameter
+  from the URL of the [Can I Use] page. A list of all available features can be
+  found at [`caniuse-lite/data/features`].
+  * `fully supports es6`: browsers with full support for specific
+    features. For example `fully supports css-grid` will omit Edge 12-15, as
+    those browser versions are marked as [having partial support].
+  * `partially supports es6-module` or `supports es6-module`:  browsers with
+    full or partial support for specific features. For example
+    `partially supports css-grid` will include Edge 12-15 support, as those
+    browser versions are marked as [having partial support].
 * `browserslist config`: the browsers defined in Browserslist config. Useful
   in Differential Serving to modify user’s config like
-  `browserslist config and partially supports es6-module`.
+  `browserslist config and fully supports es6-module`.
 * `since 2015` or `last 2 years`: all versions released since year 2015
   (also `since 2015-03` and `since 2015-03-10`).
 * `unreleased versions` or `unreleased Chrome versions`:
@@ -300,7 +302,7 @@ You can add `not ` to any query.
 [still maintained]:            https://github.com/nodejs/Release
 [Can I Use]:                   https://caniuse.com/
 [Firefox Extended Support Release]: https://support.mozilla.org/en-US/kb/choosing-firefox-update-channel
-[having partial support]: https://caniuse.com/css-grid)
+[having partial support]: https://caniuse.com/css-grid
 
 ### Grammar Definition
 
--- a/README.md
+++ b/README.md
@@ -275,8 +275,12 @@ You can specify the browser and Node.js versions by queries (case insensitive):
   `browserslist-config-mycompany` npm package.
 * `supports es6-module`: browsers with support for specific features.
   `es6-module` here is the `feat` parameter at the URL of the [Can I Use]
-  page. A list of all available features can be found at
-  [`caniuse-lite/data/features`].
+  page. This includes browsers with full or partial support for a feature.
+  A list of all available features can be found at [`caniuse-lite/data/features`].
+* `fully supports es6-module`: browsers with support for specific features.
+  `es6-module` here is the `feat` parameter at the URL of the [Can I Use]
+  page. This only includes browsers with full support for a feature.
+  A list of all available features can be found at [`caniuse-lite/data/features`].
 * `browserslist config`: the browsers defined in Browserslist config. Useful
   in Differential Serving to modify user’s config like
   `browserslist config and supports es6-module`.
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ when you add the following to `package.json`:
 
 ```json
   "browserslist": [
-    "defaults and supports es6-module",
+    "defaults and partially supports es6-module",
     "maintained node versions"
   ]
 ```
@@ -29,7 +29,7 @@ Or in `.browserslistrc` config:
 ```yaml
 # Browsers that we support
 
-defaults and supports es6-module
+defaults and partially supports es6-module
 maintained node versions
 ```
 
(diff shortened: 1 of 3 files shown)


Semigradsky:
> In an ideal world it'd feel nicer for supports to be for full support, and then we have a modifier if we want to include partial support,

👍 
From readme:
> supports es6-module: browsers with support for specific features

It doesn't look like "partial" support. Is current behaviour bug?

ai:
Can you show popular real-world example when this query will be useful?

asusvivobook4466 (approved):

asusvivobook4466 (approved):

BPScott:
Sorry for the slow reply, it's been a busy week.

> It doesn't look like "partial" support. Is current behaviour bug?

It's debatable. If I were reviewing code when supports was added then I'd consider it a bug and that I'd expect `supports` to be for full support only. However the current behaviour of including partial support has been in browserslist for so long that it's almost became relied upon behaviour and I'd be nervous about changing it outside of a major version (and a major release is too disruptive without a really really good reason so that's not happening).

"Change the behaviour of supports to omit partial support and we don't need to add any new query" would also be an acceptable solution but I doubt @ai would consider that change not disruptive enough to include in a minor/patch version.


> Can you show popular real-world example when this query will be useful?

I'm not sure I can ensure "popular", as partial support in caniuse's syntax is relatively rare these days - it was more prevalent when browsers introduced features with vendor prefixes rather than behind feature flags.

Recent features that make heavy use of partial supports are [container queries](https://caniuse.com/css-container-queries), [container style queries](https://caniuse.com/css-container-queries-style), 

`supports es6-module` is probably the most common `supports` query, and is mentioned on https://browsersl.ist. The [support for `es6-module`](https://caniuse.com/es6-module) is mostly full however Safari 10.1 and iOS 10.3 are marked as having partial support  due to them lacking support for the `nomodule` syntax. I think it is surprising that `supports es6-module` would include those two browsers.

If users wished to use `nomodule` as a way of doing differential serving - sending legacy js to older browsers and  a build targeting `supports es6-module` to newer browsers with: `<script nomodule src="/dist/index.legacy.js"></script> <script type="module" src="/dist/index.esm.js"></script>` users would find that their esm build would contain transforms and polyfills for Safari 10.1 and iOS 10.3  even through this code would never be served to those browsers thanks to the lack of `nomodule` support.


Currently [`defaults and supports es6`](https://browsersl.ist/#q=defaults+and+supports+es6%0A) would include KaiOS 2.5, which only has partial es6 support which seems feels unexpected at first glance.


My particular use-case is for creating builds for long-tail legacy browsers, I'd want to use `fully supports es6` as a mechanism to select all browsers with es6 support, `supports es6` contains a multitude of browsers that have partial support for es6 and using it as a target would result in the inclusion of several undesired transforms and polyfills.

ai:
Yes, these use cases look legit. I will think about adding it.

README.md
```diff
@@ -275,8 +275,12 @@ You can specify the browser and Node.js versions by queries (case insensitive):
   `browserslist-config-mycompany` npm package.
 * `supports es6-module`: browsers with support for specific features.
   `es6-module` here is the `feat` parameter at the URL of the [Can I Use]
-  page. A list of all available features can be found at
-  [`caniuse-lite/data/features`].
+  page. This includes browsers with full or partial support for a feature.
+  A list of all available features can be found at [`caniuse-lite/data/features`].
+* `fully supports es6-module`: browsers with support for specific features.
```
ai, line 280: Can you improve docs to clarify the difference in query description?

README.md
```diff
@@ -275,8 +275,12 @@ You can specify the browser and Node.js versions by queries (case insensitive):
   `browserslist-config-mycompany` npm package.
 * `supports es6-module`: browsers with support for specific features.
   `es6-module` here is the `feat` parameter at the URL of the [Can I Use]
-  page. A list of all available features can be found at
-  [`caniuse-lite/data/features`].
+  page. This includes browsers with full or partial support for a feature.
```
ai, line 278: Can you also send PR to [our website](https://browsersl.ist/) repo: https://github.com/browserslist/browsersl.ist/blob/main/client/index.html

README.md
BPScott: I can try! 

I already mention that `supports` is for "full or partial support", while "fully supports" says "This only includes browsers with full support for a feature."

I could give an example with something like `supports es6` vs `fully supports es6`.

Did you have any thoughts on what'd be a good way to clarify the difference?

README.md
BPScott: Will do!

ai: > Should I push for using fully supports as a default here to match the intent of that text? 

Yes, we can do it

README.md
ai: > Did you have any thoughts on what'd be a good way to clarify the difference?

I am not native speaker and docs is harder for me 😅

README.md
```diff
@@ -273,13 +273,18 @@ You can specify the browser and Node.js versions by queries (case insensitive):
...
+  A list of all available features can be found at [`caniuse-lite/data/features`].
+* `partially supports es6-module`: browsers with full or partial support for specific features.
+  `es6-module` here is the `feat` parameter at the URL of the [Can I Use]
+  page. For example `fully supports css-grid` will include Edge 12-15 support, as those browser versions are marked as [having partial support].
+  A list of all available features can be found at [`caniuse-lite/data/features`].
+* `supports es6-module`: browsers with full or partial support for specific features. An alias for `partially supports es6-module`.
 * `browserslist config`: the browsers defined in Browserslist config. Useful
   in Differential Serving to modify user’s config like
-  `browserslist config and supports es6-module`.
+  `browserslist config and partially supports es6-module`.
```
ai, line 287:
```suggestion
  `browserslist config and partially fully es6-module`.
```

BPScott:
I've pushed up a new commit, to try to improve the documentation by adding an example referencing `css-grid` (which is a mentioned example query on https://browsersl.ist/.

When thinking about documentation and how to teach this I felt that the difference between `supports` and `fully supports` wasn't super obvious.

So try to solve this I've added a `partially supports` query that acts the same as `supports`. The documentation describes the difference between `fully supports` and `partially supports`, and then notes that `supports` is an alias for `partially supports`. In the future, Browserslist v5 could remove `supports` or make it be an alias for `fully supports`.


---

I've moved the existing examples of `supports` queries to use `partially supports`. I'm tempted to say these should be `fully supports`, @ai do you have a preference?

README.md
```diff
@@ -273,13 +273,18 @@ You can specify the browser and Node.js versions by queries (case insensitive):
...
 * `extends browserslist-config-mycompany`: take queries from
   `browserslist-config-mycompany` npm package.
-* `supports es6-module`: browsers with support for specific features.
+* `fully supports es6-module`: browsers with full support for specific features.
   `es6-module` here is the `feat` parameter at the URL of the [Can I Use]
-  page. A list of all available features can be found at
-  [`caniuse-lite/data/features`].
+  page. For example `fully supports css-grid` will omit Edge 12-15, as those browser versions are marked as [having partial support].
+  A list of all available features can be found at [`caniuse-lite/data/features`].
+* `partially supports es6-module`: browsers with full or partial support for specific features.
```
ai, line 280: We need to mention old query name too

README.md
```diff
@@ -273,13 +273,18 @@ You can specify the browser and Node.js versions by queries (case insensitive):
...
+  `es6-module` here is the `feat` parameter at the URL of the [Can I Use]
```
ai, line 281: Let’s avoid `es-module` dublicate here

README.md
```diff
@@ -273,13 +273,18 @@ You can specify the browser and Node.js versions by queries (case insensitive):
     to PhantomJS runtime.
 * `extends browserslist-config-mycompany`: take queries from
   `browserslist-config-mycompany` npm package.
-* `supports es6-module`: browsers with support for specific features.
+* `fully supports es6-module`: browsers with full support for specific features.
   `es6-module` here is the `feat` parameter at the URL of the [Can I Use]
-  page. A list of all available features can be found at
-  [`caniuse-lite/data/features`].
+  page. For example `fully supports css-grid` will omit Edge 12-15, as those browser versions are marked as [having partial support].
```
ai, line 278: We need to keep 80 symbols in a line

ai:
> I'm tempted to say these should be fully supports, @ai do you have a preference?

Let’s move to `fully` in examples

README.md
BPScott: The next bullet point in this list mentions `supports es6-module`. What if I remove the dedicated line below and update this bullet point to be:

> `partially supports es6-module` or `supports es6-module`: browsers with full or partial support for specific features. 

Does that sound good?

ai: Yes, sounds good

BPScott:
Let me know when you're you've got no more feedback, and I can squash these commits down prior to merge.

README.md
BPScott: PR to browsersl.ist added: https://github.com/browserslist/browsersl.ist/pull/522

ai:
Thanks. Released in 4.22.

PR #791, opened 2023-09-21
Add update-browserslist-db mention on --update-db call and deprecation hint.

konclave:
The --update-db key is missing in the usage hint printed to the terminal when CLI is called with the --help key.

### DIFF
--- a/cli.js
+++ b/cli.js
@@ -20,8 +20,7 @@ var USAGE =
   '  npx browserslist --env="environment name defined in config"\n' +
   '  npx browserslist --stats="path/to/browserlist/stats/file"\n' +
   '  npx browserslist --mobile-to-desktop\n' +
-  '  npx browserslist --ignore-unknown-versions\n' +
-  '  npx browserslist --update-db\n'
+  '  npx browserslist --ignore-unknown-versions\n'
 
 function isArg(arg) {
   return args.some(function (str) {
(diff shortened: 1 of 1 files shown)


ai:
We don’t have `--update-db` here anymore. It is now just an alias for external tool `npx update-browserslist-db`.

This is why we removed it from docs.

Do you know how we can mention it in another way?

konclave:
But `update-browserslist-db` prints to console warning: `npx browserslist@latest --update-db`, and I was pretty surprised when I ran `npx browserslist@latest --help` and didn't see this key. 

As I understand, `--update-db` is kept for the legacy support, right? Then, probably, it is worth replacing the console hint from `update-browserslist-db` with `run npx update-browserslist-db ...`?

ai:
> prints to console warning: npx browserslist@latest --update-db

I fixed the warning https://github.com/browserslist/update-db/commit/3fff4d6cd277e97f50166fbdf6262fe82c92518a

But are you sure that this warning was not from Browserslist?

konclave:
About 75% sure =). I did the repository search in GitHub and haven't found any string containing `npx browserslist@latest --update-db`. And the only place with just `--update-db` is cli.js.

ai:
Since we fixed warning there. Can we close it or do you have maybe another idea what to add to docs here?

konclave:
Probably, add a hint when browserslist is called with `--update-db` that tells that `--update-db` is deprecated and `npx update-browserslist-db` should be used instead.

That's the last idea I have.

ai:
Yes, I like it. Can you send PR? :)

konclave:
Sure

konclave:
done

cli.js
```diff
@@ -38,6 +38,8 @@ if (isArg('--help') || isArg('-h')) {
 } else if (isArg('--version') || isArg('-v')) {
   process.stdout.write('browserslist ' + pkg.version + '\n')
 } else if (isArg('--update-db')) {
+  process.stdout.write('The --update-db command is deprecated. Please use npx update-browserslist-db@latest instead.\n')
+  process.stdout.write('Browserslist DB update will still be made.\n')
   /* c8 ignore next 3 */
```
ai, line 43: Can you just move this lines above and increase 3 to 5

konclave: As you merged it, I won't make any following changes.

PR #816, opened 2024-01-26
Updated SCOPED_CONFIG__PATTERN regex 

boucodes:
Updated regex to have the option of adding an extension after @compan…


ai:
Thanks

PR #819, opened 2024-02-12
feat: add BROWSERSLIST_ROOT_PATH

### LINKED ISSUE #813: Browserslist assumes it has read access to all directories in path
Browserslist throws a fatal error if it fails to read a file or directory while searching for config files on the ancestors of the path. From the Readme:

"Browserslist will check config in every directory in path."

This causes problems if running this in Deno with restrictive permissions. A typical usage will be to set `--allow-read=.`. However this causes BL to throw an error as it tries to find configs up the path.

BL does the same thing with usage stats. It goes up the path and throws a fatal error if denied read access.

I realize I can get around this problem by explicitly passing the config and usage_stats files as environment variables. That's fine for the config, but it's awkward for usage stats: I have to create a valid json file even if I don't have any stats to work with.

Furthermore it's a frustrating developer experience to have a script throw an error trying to read something that doesn't even exist (a bl config up the path) given that it will carry on without even a warning if it it successfully determines there is nothing there.

I wonder if it would be possible to give users the ability to stop this behavior, or control it? One idea:

`BROWSERSLIST_ROOT_PATH`: browserslist will check for config or usage_data files up to and including this path.

Note that while my issue is entirely due to my choice of using Deno and with restrictive permissions, Node too is working on a permissions system, so this issue may be relevant for node users someday too.

Thank you!

teleclimber:
- check that loc is inside BROWSERSLIST_ROOT_PATH in eachParent
- add tests
- update Readme
- fixes #813

### DIFF
--- a/README.md
+++ b/README.md
@@ -683,6 +683,12 @@ with [environment variables]:
   BROWSERSLIST_DANGEROUS_EXTEND=1 npx webpack
    ```
 
+* `BROWSERSLIST_ROOT_PATH` to prevent reading files above this path.
+
+   ```sh
+  BROWSERSLIST_ROOT_PATH=. npx webpack
+   ```
+
 [environment variables]: https://en.wikipedia.org/wiki/Environment_variable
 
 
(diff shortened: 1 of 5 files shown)


ai:
Looks great. I am going to release it tomorrow.

ai:
Thanks. Released in 4.23.

PR #844, opened 2024-09-24
Introduce `browserslist.findConfigFile` API

JLHwung:
In this PR we introduce a new API `findConfigFile` that returns the absolute path of the loaded browserslist config: one of `package.json`, `.browserslistrc` or `browserslist`.

The motivation of this PR is to pass `@babel/preset-env` the loaded browserslist config files, so that the env preset can then call the Babel plugin `api.addDependency` to have other tools (such as `babel-loader` and `@rollup/plugin-babel`) to recognize the browserslist config as a build dependency.

See also https://github.com/babel/babel-loader/issues/690

### DIFF
--- a/browser.js
+++ b/browser.js
@@ -44,6 +44,8 @@ module.exports = {
 
   findConfig: noop,
 
+  findConfigFile: noop,
+
   clearCaches: noop,
 
   oldDataWarning: noop,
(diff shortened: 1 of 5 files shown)


ai:
Thanks. Released in 4.24.0.

PR #849, opened 2024-10-21
Updated oldDataWarning function to display how many months old caniuse-lite is.

mezhnin:


node.js
```diff
@@ -419,8 +429,10 @@ module.exports = {
     var halfYearAgo = Date.now() - TIME_TO_UPDATE_CANIUSE
 
     if (latest !== 0 && latest < halfYearAgo) {
+      var monthPassed = getMonthsPassed(latest)
+
       console.warn(
-        'Browserslist: caniuse-lite is outdated. Please run:\n' +
+        'Browserslist: caniuse-lite is '+ monthPassed +' month old. Please run:\n' +
```
ai, line 435: We need plural form `month`/`months` and tests

mezhnin: ok, let me fix it

node.js
```diff
@@ -419,8 +429,10 @@ module.exports = {
     var halfYearAgo = Date.now() - TIME_TO_UPDATE_CANIUSE
```
ai, line 429: Why you need `halfYearAgo` if you already calculated month?

node.js
mezhnin: fixed

node.js
mezhnin: fixed

PR #850, opened 2024-10-21
Display time since the last update of caniuse-lite

echo-vladimir:
The PR request specifies a warning message for outdated caniuse-lite data, detailing the exact time since the last update

Changes:
- Implemented a warning that notifies users when caniuse-lite is outdated
- The warning now specifies the time (in months) since the last browser release related to the data
- Updated the relevant tests to verify the new functionality

### DIFF
--- a/node.js
+++ b/node.js
@@ -416,13 +416,13 @@ module.exports = {
     if (process.env.BROWSERSLIST_IGNORE_OLD_DATA) return
 
     var latest = latestReleaseTime(agentsObj)
-    var halfYearAgo = Date.now() - TIME_TO_UPDATE_CANIUSE
+    var currentTime = Date.now()
 
-    if (latest !== 0 && latest < halfYearAgo) {
+    if (latest !== 0 && latest < currentTime - TIME_TO_UPDATE_CANIUSE) {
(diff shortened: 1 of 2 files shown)


node.js
```diff
@@ -419,11 +419,18 @@ module.exports = {
     var halfYearAgo = Date.now() - TIME_TO_UPDATE_CANIUSE
```
ai, line 419: You can use the same month for `halfYearAgo` check

node.js
```diff
@@ -416,14 +416,21 @@ module.exports = {
     if (process.env.BROWSERSLIST_IGNORE_OLD_DATA) return
 
     var latest = latestReleaseTime(agentsObj)
-    var halfYearAgo = Date.now() - TIME_TO_UPDATE_CANIUSE
+    var currentTime = Date.now()
+
+    if (latest !== 0 && latest < currentTime - TIME_TO_UPDATE_CANIUSE) {
```
ai, line 421: `1000 * 60 * 60 * 24 * 30` math is dublicated

ai: You can get month from last update and then compare it with `> 6` and not use `TIME_TO_UPDATE_CANIUSE`

Issue #859, opened 2024-12-14
Config resolution performance


thoughtspile:
Hi there! We love browserslist and the overall postcss stack, but we've recently run into some performace issues with `vite -> postcss -> autoprefixer -> browserslist`. We maintain a mid-sized svelte app with ~2K CSS files / components, and I've noticed browserslist takes up ~2.5s of the build time. Not a deal breaker, but I dislike waste, so I went to investigate. I know you take dev setup performance very seriously, so I thought I'd share my concerns.

<img width="1075" alt="Снимок экрана 2024-12-14 в 15 41 41" src="https://github.com/user-attachments/assets/80519c2f-1771-4712-9fb3-767d58ade28a" />

A sizable portion of this time (1s / 40%) is spent in [parsePackage.](https://github.com/browserslist/browserslist/blob/0639f5484b46b6c8608eab13a22310e9922feb96/node.js#L105) I can see why, as our UI library has a gigantic 16K-line / 1Mb `package.json`, which is read 250 times. Reading the root project's package 2K times adds another 170ms. We _could_ work around this by e.g. setting `process.env.BROWSERSLIST_CONFIG`, but it would be cool to avoid this problem by default.

So, first things, first. __Is reading browserslist configs living in `node_modules` intended?__ Personally, I wouldn't expect stray configs in published libraries to take precedence over the root config of the end project. 

Next, caching. I see findConfig [caches resolved config](https://github.com/browserslist/browserslist/blob/0639f5484b46b6c8608eab13a22310e9922feb96/node.js#L405) for all the directories between the css file and the resolved config. However, in our case the cache does not work:
1. In svelte + vite (and, possibly, other SFC frameworks like Vue), `file` points to a virtual module (e.g. `/Users/vklepov/Documents/project/src/Component.svelte?svelte&type=style&lang.css`), so `isFile` returns `false`, and unless we've hit the exact same component in the past, we go to [findConfigFile branch](https://github.com/browserslist/browserslist/blob/0639f5484b46b6c8608eab13a22310e9922feb96/node.js#L397)
2. `findConfigFile` is not aware of `configCache`, so we end up reading configs over and over again.

I don't see a simple fix for fileness check on virtual modules, but attaching `findConfigFile` to `configCache` should be quite easy, and greatly reduce FS operations — we'd only traverse `eachParent` until we hit a previously visited common ancestor directory.

[...]

PR #862, opened 2024-12-16
Remove path from cache key for path-independent queries

### LINKED ISSUE #860: opts.path interferes with resolved query caching
When debugging our build, I've noticed that even after forcing `BROWSERSLIST_CONFIG` as a workaround for #859 , browserslist still takes ~1s to run. About half that time is spent on [repeatedly resolving](https://github.com/browserslist/browserslist/blob/0639f5484b46b6c8608eab13a22310e9922feb96/index.js#L424) our single browserslist config into browsers:

Снимок экрана 2024-12-14 в 17 28 08

From the looks of it, the resolution cache key includes `opts.path`, which, in our case (vite -> autoprefixer -> browserslist) equals CSS file path:

https://github.com/browserslist/browserslist/blob/0639f5484b46b6c8608eab13a22310e9922feb96/index.js#L404C1-L421C52

However, nothing in `resolve` seems to reference `path` (or any `context` field other than `mobileToDesktop`, for that matter. 

Could we maybe exclude `path` from `cacheKey`? Together with the #859 workaround, this brings browserslist time to a reasonable 400ms:

Снимок экрана 2024-12-14 в 17 35 09

thoughtspile:
fixes #860

Strawman: remove `path` from `context` / `cacheKey` when no `QUERIES` depend on path. `hasPathAwareQueries` checks for `browserslist_config` and `extends` directives, uses ad-hoc regular expression without `^/$` to avoid calling `parse`

### DIFF
--- a/index.js
+++ b/index.js
@@ -6,7 +6,7 @@ var path = require('path')
 
 var BrowserslistError = require('./error')
 var env = require('./node')
-var parse = require('./parse') // Will load browser.js in webpack
+var parseWithoutCache = require('./parse') // Will load browser.js in webpack
 
 var YEAR = 365.259641 * 24 * 60 * 60 * 1000
 var ANDROID_EVERGREEN_FIRST = '37'
@@ -319,7 +319,7 @@ function isSupported(flags, withPartial) {
 }
 
 function resolve(queries, context) {
-  return parse(QUERIES, queries).reduce(function (result, node, index) {
+  return parseQueries(queries).reduce(function (result, node, index) {
     if (node.not && index === 0) {
       throw new BrowserslistError(
         'Write any browsers query (for instance, `defaults`) ' +
@@ -395,13 +395,14 @@ function checkQueries(queries) {
 }
 
 var cache = {}
+var parseCache = {}
 
 function browserslist(queries, opts) {
   opts = prepareOpts(opts)
   queries = prepareQueries(queries, opts)
   checkQueries(queries)
 
-  var needsPath = parse(QUERIES, queries).some(function (node) {
+  var needsPath = parseQueries(queries).some(function (node) {
     return QUERIES[node.type].needsPath
   })
   var context = {
@@ -446,11 +447,21 @@ function browserslist(queries, opts) {
   return result
 }
 
+function parseQueries(queries) {
+  var cacheKey = JSON.stringify(queries)
+  if (cacheKey in parseCache) return parseCache[cacheKey]
+  var result = parseWithoutCache(QUERIES, queries)
+  if (!env.env.BROWSERSLIST_DISABLE_CACHE) {
+    parseCache[cacheKey] = result
+  }
+  return result
+}
+
 browserslist.parse = function (queries, opts) {
   opts = prepareOpts(opts)
   queries = prepareQueries(queries, opts)
   checkQueries(queries)
-  return parse(QUERIES, queries)
+  return parseQueries(queries)
 }
 
 // Will be filled by Can I Use data below
--- a/index.js
+++ b/index.js
@@ -401,12 +401,17 @@ function browserslist(queries, opts) {
   queries = prepareQueries(queries, opts)
   checkQueries(queries)
 
+  var needsPath = parse(QUERIES, queries).some(function (node) {
+    return QUERIES[node.type].needsPath
+  })
   var context = {
     ignoreUnknownVersions: opts.ignoreUnknownVersions,
     dangerousExtend: opts.dangerousExtend,
     mobileToDesktop: opts.mobileToDesktop,
-    path: opts.path,
     env: opts.env
+  }
+  if (needsPath) {
+    context.path = opts.path
   }
 
   env.oldDataWarning(browserslist.data)
@@ -1133,6 +1138,7 @@ var QUERIES = {
   browserslist_config: {
     matches: [],
     regexp: /^browserslist config$/i,
+    needsPath: true,
     select: function (context) {
       return browserslist(undefined, context)
     }
@@ -1140,6 +1146,7 @@ var QUERIES = {
   extends: {
     matches: ['config'],
     regexp: /^extends (.+)$/i,
+    needsPath: true,
     select: function (context, node) {
       return resolve(env.loadQueries(context, node.config), context)
     }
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -418,6 +417,10 @@ function browserslist(queries, opts) {
     }
   }
 
+  if (hasPathAwareQueries(queries)) {
+    context.path = opts.path;
```
ai, line 421: This fix is too hacky. 1 year later I will forget this hack and will spend a night trying to find why `context` is different for some cases.

We should keep `context` the same, but change cache algorithm instead.

thoughtspile: Are you suggesting removing path during cacheKey generation? I thought it would be more risky since accessing context.path from another query would appear to work initially, but fail subtly due to caching, as opposed to failing loudly due to missing path in the current fix.

The clean solution would probably be: 
1. Add explicit dependencies, eg ['path', 'mobileToDesktop'] to every query
2. Do a parse() pass on query list to split on logical operators
3. Form context with discovered dependencies
But it would be less efficient due to parsing, and probably not that useful as the query set is probably stable by now. 

If you have some entirely different caching algorithm in mind, I'd be glad to add it, but I don't see what it could be.

ai: I suggest to *not* remove `path` from `context` and instead change the code which use cache.

> The clean solution would probably be:

Also, a way but maybe too complex.

Can you show me the exact place where you want to change cache logic?

thoughtspile: If we attack from a different angle, we can normalize `path` to reduce cardinality — even path-dependent queries don't seem to need the precise current file path:

- `browserslist_config` needs the closest directory with browserslist config
- `extends` needs the closest `package.json` (unless using `dangerousExtend`)

So we could replace `path` with something like

```js
var context = {
  extendsRoot: dangerousExtend ? opts.path : findClosestPackageJson(opts.path),
  parentConfig: env.findConfigFile(opts.path),
  ...
};
```

In this case, normalizing the path involves FS, which brings us back to #859

ai: Hm, interesting.

Let’s finish with #864 and then will try to do this.

thoughtspile: before we go all crazy — pls take a look at the updated option, is this too hacky still?

index.js
```diff
@@ -441,11 +447,21 @@ function browserslist(queries, opts) {
   return result
 }
 
+function parseQueries(queries) {
+  var cacheKey = JSON.stringify(queries)
```
ai, line 451: I think a better way is to use `configPath` as cache key and not config content.

With `configPath` as cache key we don’t need to parse config, etc.

thoughtspile: There is not always a `configPath`, because we can get config from `opts.queries`, `env.BROWSERSLIST`, or the fallback to `browserslist.defaults|

Besides, there's no repeated parsing, because `parsePackageOrReadConfig` now had its own cache

ai: Hm. Yes, you are right.

In this case, the path is great. Let’s avoid `ctx.path` changes and we are ready to merge.

thoughtspile: Sorry, still don't get what you mean by "avoid `ctx.path` changes". Something like this?

```diff
-var cacheKey = JSON.stringify([queries, context])
+var cacheKey = JSON.stringify([queries, {
+  ignoreUnknownVersions: context.ignoreUnknownVersions,
+  dangerousExtend: context.dangerousExtend,
+  mobileToDesktop: context.mobileToDesktop,
+  env: context.env,
+  path: needsPath ? context.path : undefined
+}])
```

ai: https://github.com/browserslist/browserslist/pull/862#discussion_r1888380971

ai: For instance, add `ctx.cachePath` or something like this and check every time when we use `ctx.path`

thoughtspile: I only see two ways to, at the same time:

1. Keep `path` in context
2. Remove `path` from cache key
3. Avoid cloning `context` in multiple places, which is verbose and error-prone in es5

One is the `definePropery` trick I just pushed, the other is a sprinkle of TS to enforce `usesPath` on queries that need `{ path: string }`. I'm probably missing the point, but see if you find any of the options suitable.

thoughtspile: Hey @ai, could you take a look?

ai: Now I get it why you deleted `ctx.path`. I initially thought that it is to disable cache, but it is because some query was not subscribed to the path?

index.js
```diff
@@ -395,19 +395,27 @@ function checkQueries(queries) {
...
+    return QUERIES[node.type].needsPath
+  })
   var context = {
     ignoreUnknownVersions: opts.ignoreUnknownVersions,
     dangerousExtend: opts.dangerousExtend,
     mobileToDesktop: opts.mobileToDesktop,
-    path: opts.path,
     env: opts.env
   }
+  Object.defineProperty(context, 'path', needsPath ? { value: opts.path } : {
```
ai, line 414: OK, when I finaylly get why we removed `ctx.path` let’s remove it :D But add a comment like

```
// Removing to avoid using ctx.path without marking query
```

thoughtspile: ✅

index.js
thoughtspile: OK, since I still don't understand the confusion, here's the most detailed and boring exlainer to avoid any miscommunication:

1. `context` object is passed to all `QUERIES[*].select` _and_ used as the cache key
2. `context` contains `opts.path`, which is the path of the file currently processed
3. Putting `path` in cache key causes resolve cache to become useless for processing multiple files
4. Only two queries, `browserslist config` and `extends` actually use `path`

Therefore, we remove `path` from `context` unsell config contains any of the two path-dependent queries. 

We accept two risks:

1. Reading `context.path` from a query not marked with `needsPath` will fail, and probably cause debug trouble
2. Resolving configs with `browserslist config` or `extends` queries will be de-optimized and mostly bypass cache.

I could create an issue for risk 2 if you want, fixing (non-`dangerous`) `extends` shouldn't be too hard.

ai: After the holiday I will do I final review for this PR.

You can add “non-dangerous extends” support to this PR if you have a free time before it.

Dangerous extends is much rare case, so it is OK to keep it de-optimized. Non-dangerous extends is rare too, but I know that some people use it.

ai:
New cache released in 4.24.4. Thanks!

thoughtspile:
Cool, browserslist time goes from 2.55s to 175ms for our project! Thanks for releasing, and please do reach out in case I broke something.

PR #864, opened 2024-12-21
Optimize FS cache

### LINKED ISSUE #859: Config resolution performance

thoughtspile:
Fixes #859 

Bundled a few things together:
1. Move cache from findConfigFile to eachParent, so that we
    1. Use cache if _any_ parent has been cached, not just on exact match of the first file / directory passed to `findConfig`
    2. Avoid a second `eachParent` pass to populate cache
    3. Avoid trouble when cache value exists, but is undefined, and won't terminate `eachParent` when returned from the callback
2. Add cache for `parsePackage` / `readConfig`, since `findConfigFile` caches paths, not content
3. Replaced `isFile` check in `eachParent` with explicit `isDirectory` check to optimize for virtual modules (path provided does not exist, is considered a directory, yields 3 fs checks for all kinds of config)
4. Added cache to `getStat` now that it's so easy with eachParent supporting cache
5. Removed `filenessCache` since `isFile` is only called within `eachParent`, and both `eachParent` calls now have explicit caches and will be called once for each directory

Tell me if I\ve gone too far, steps 1+2 already work pretty well.

### DIFF
--- a/node.js
+++ b/node.js
@@ -387,7 +387,7 @@ module.exports = {
   },
 
   findConfigFile: function findConfigFile(from) {
-    var resolved = eachParent(from, function (dir) {
+    return eachParent(from, function (dir) {
       var config = path.join(dir, 'browserslist')
       var pkg = path.join(dir, 'package.json')
       var rc = path.join(dir, '.browserslistrc')
@@ -426,8 +426,6 @@ module.exports = {
     }, {
       cache: configPathCache
     })
-
-    return resolved
   },
 
   findConfig: function findConfig(from) {
--- a/node.js
+++ b/node.js
@@ -14,7 +14,9 @@ var FORMAT =
 
 var dataTimeChecked = false
 var filenessCache = {}
-var configCache = {}
+var configPathCache = {}
+var parseConfigCache = {}
+
 function checkExtend(name) {
   var use = ' Use `dangerousExtend` option to disable.'
   if (!CONFIG_PATTERN.test(name) && !SCOPED_CONFIG__PATTERN.test(name)) {
@@ -103,6 +105,10 @@ function pickEnv(config, opts) {
 }
 
 function parsePackage(file) {
+  if (file in parseConfigCache) {
+    return parseConfigCache[file]
+  }
+  
   var text = fs
     .readFileSync(file)
     .toString()
@@ -123,6 +129,10 @@ function parsePackage(file) {
   }
   for (var i in list) {
     check(list[i])
+  }
+
+  if (!process.env.BROWSERSLIST_DISABLE_CACHE) {
+    parseConfigCache[file] = list
   }
 
   return list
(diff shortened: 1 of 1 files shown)


ai:
Can you resolve the confict so we can run CI?

thoughtspile:
there you go

node.js
```diff
@@ -35,25 +37,44 @@ function checkExtend(name) {
...
+  return fs.existsSync(file) && fs.statSync(file).isFile()
+}
+function isDirectory(dir) {
+  return fs.existsSync(dir) && fs.statSync(dir).isDirectory()
 }
 
-function eachParent(file, callback) {
-  var dir = isFile(file) ? path.dirname(file) : file
-  var loc = path.resolve(dir)
+function eachParent(file, callback, options) {
```
ai, line 46: Let’s just pass `cache` to last argument since it is private function

node.js
```diff
@@ -35,25 +37,44 @@ function checkExtend(name) {
...
+  var loc = path.resolve(file)
+  var cache = options ? options.cache : false
+  var traceback = []
```
ai, line 49: Not clear name. Let’s rename for instance to `pathsForCacheResult`

juana-18:
new

PR #887, opened 2025-04-19
Protect against prototype pollution in loadCountry, normalizeStats, and loadFeature

MegaManSec:
Also harden against prototype pollution elsewhere, and remove an invalid extra parameter in a function call.

Also use a stack-based approach to avoid a potential stack overflow (which results in a throw) in flatten.

### DIFF
--- a/index.js
+++ b/index.js
@@ -254,7 +254,7 @@ function normalizeAndroidVersions(androidVersions, chromeVersions) {
 }
 
 function copyObject(obj) {
-  var copy = {}
+  var copy = Object.create(null)
   for (var key in obj) {
     copy[key] = obj[key]
   }
@@ -517,13 +517,13 @@ browserslist.coverage = function (browsers, stats) {
   if (typeof stats === 'undefined') {
     data = browserslist.usage.global
   } else if (stats === 'my stats') {
-    var opts = {}
+    var opts = Object.create(null)
     opts.path = path.resolve ? path.resolve('.') : '.'
     var customStats = env.getStat(opts)
     if (!customStats) {
       throw new BrowserslistError('Custom usage statistics was not provided')
     }
-    data = {}
+    data = Object.create(null)
     for (var browser in customStats) {
       fillUsage(data, browser, customStats[browser])
     }
@@ -539,7 +539,7 @@ browserslist.coverage = function (browsers, stats) {
     if ('dataByBrowser' in stats) {
       stats = stats.dataByBrowser
     }
-    data = {}
+    data = Object.create(null)
     for (var name in stats) {
       for (var version in stats[name]) {
         data[name + ' ' + version] = stats[name][version]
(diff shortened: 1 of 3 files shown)


ai:
1. How do you see someone is theoretically using this attack?
2. Does it protect from any harm which could be done by prototype pollution? Or is it just automatic `{}` → `Object.create(null)` replacement?

MegaManSec:
Hey,

1. Generally speaking, within a larger build system/eco-system where `caniuse-lite` `js` files are writable by lesser-privileged users than developers (or, in a build system/eco-system which _code_ review happens, but changes to `js` files are not reviewed).
2. No, not an automatic replacement. There are many other places I would have liked to have changed, including some checks on the keys, but I don't have the time to create testcases to ensure 100% code coverage which is necessary. As for what it protects, well, it depends. For example:
   ```
      usage[country] = {}
      for (var i in usageData) {
        for (var j in usageData[i]) {
          usage[country][i + ' ' + j] = usageData[i][j]
        }
      }
   ```
   In this case, consider where `country = __proto__` and `i` and `j` are arbitrary, resulting in some property on the prototype being set with arbitrary data. This is generally not as dangerous as it could be, because there is the space, and defining some property on the prototype is limited to properties with spaces in their name. The other place, however, isn't limited to that:
   ```
    var stats = feature(compressed).stats
    features[name] = {}
    for (var i in stats) {
      features[name][i] = {}
      for (var j in stats[i]) {
        features[name][i][j] = stats[i][j]
      }
    }
   ```

Note: just thought I'd mention that this PR will not be followed by any CVE request or anything, it's just some internal hardening that we've created and I always try to upstream some of the changes where I can (aka: not farming for security issues automatically to waste everybody's time).

ai:
These changes will increase maintenance cost for me a lot so I prefer to increase code complety only if there are objective benefits for users.

I do not see those benefits yet because I can't imagine an attack vector (user who can write to `caniuse-lite` can also write to Browserslist code).

It looks like security theater for now without extra proofs

MegaManSec:
Hey,

>user who can write to caniuse-lite can also write to Browserslist code

This isn't true if using (for example) GitHub's branch protection, requiring code reviews from code owners, with a CODEOWNERS such as:

```
# No required review
*.json

# Requires approval
*.js @org/js-maintainers
```

Since `json` files are not expected to directly influence _unexpected_ variables, allowing lesser-privileged users (think: customer service, new hires, whatever) to edit them does not normally pose a risk. At the moment, a benign json file can potentially be turned into a malicious one. This is not expected, and is simply what these changes make.

Again, if it's preferred not to merge them upstream, that's fine.

Cheers,

ai:
How JSON changes can pollute `Object`?

MegaManSec:
```js
function example(data) {
  const object = {};
  for (const outerKey in data) {
    const innerKeys = Object.keys(data[outerKey]);
    for (const innerKey of innerKeys) {
      object[outerKey][innerKey] = data[outerKey][innerKey];
    }
  }
  return object;
}

const malicious = JSON.parse(`{
  "__proto__": {
    "isAdmin": true
  }
}`);

a = example(malicious);
console.log({}.isAdmin);
```

```
$ node pp2.js 
true
```

MegaManSec:
Yeah ok sorry, the changes in `index.js` are a big extraneous and were not meant to be included in this PR (only the `browserslist.versionAliases[name] = {}` line). I was talking about the changes in `node.js`, not realizing there were other changes here.

PR #909, opened 2025-11-08
feat: allow esr alias in firefox version ranges

SethFalco:
As of this message, Firefox ESR is an alias for Firefox 140. However, as Firefox ESR is treated as its own browser, it's not possible to use this in version ranges as far as I know.

This works already, which is effectively the same as Firefox 140:

```
Firefox ESR
```

This does not work, which I'd expect it to be the same as Firefox >= 140:

```
Firefox >= ESR
```

Would you allow `ESR` to be an acceptable alias in `browser_ray` as well? Then `ESR` will be mapped to Firefox 140 when both specifying the browser but also in version ranges.

## Use Case

For [SVGO.dev](https://svgo.dev), as the website is aimed at developers, I'm looking at only supporting modern browsers.

I'm thinking a browserslist query like this, which has 86.9% global coverage, or 93.85~% coverage from my stats:

```
baseline widely available and not Firefox >= 0
Firefox >= ESR
not dead
```

### DIFF
--- a/index.js
+++ b/index.js
@@ -12,6 +12,7 @@ var parseWithoutCache = require('./parse') // Will load browser.js in webpack
 var YEAR = 365.259641 * 24 * 60 * 60 * 1000
 var ANDROID_EVERGREEN_FIRST = '37'
 var OP_MOB_BLINK_FIRST = 14
+var FIREFOX_ESR_VERSION = '140'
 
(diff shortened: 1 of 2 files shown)


ai:
Thanks. Released in 4.48.0.

PR #912, opened 2025-11-23
Add GitHub Actions workflow for npm package publishing

Kelleretoro:
This workflow runs tests and publishes a Node.js package to GitHub Packages upon release creation.


ai:
We don’t need build and we publish it in more secure way with Yubikey

PR #939, opened 2026-08-05
fix: support "including kaios" without downstream

Jaybhade:
Rewritten to make the query work instead of rejecting it, per review.

## The bug

`baseline … including kaios` **without** `with downstream` never returns KaiOS.
What it does instead depends on which `baseline-browser-mapping` you resolved:

| `baseline-browser-mapping` | `baseline widely available including kaios` |
| --- | --- |
| `2.10.44` (what `pnpm-lock.yaml` pins, so what CI runs) | `process.exit(1)` |
| `2.11.12` (what `^2.10.44` resolves to today) | silently identical to `baseline widely available` |

`select()` forwards `includeKaiOS` and `includeDownstreamBrowsers`
independently, and `baseline-browser-mapping` refuses that pair. Up to 2.10.44
it refused with `console.log(new Error(…))` followed by `process.exit(1)` — not
a throw, so nothing was catchable and `suppressWarnings: true` did not help.
Any tool resolving the config (webpack, Babel, Autoprefixer, Stylelint) just
died with exit 1 and an error on **stdout**. 2.11.x turned that into a real
`Error`, but only in `getAllVersions`/`getTimeline`; `getCompatibleVersions`,
the one we call, now drops KaiOS without a word.

So on current npm the query is a silent no-op, and on the locked version it is
fatal. Both are wrong, and a user pinning either can hit either.

## The fix

KaiOS is downstream *in `baseline-browser-mapping`'s model*, but from
Browserslist's side there is nothing incoherent about "core browsers plus
KaiOS" — the mapping just has no way to express it. So ask for the whole
downstream set, keep KaiOS out of it, and merge that onto the core set:

```js
options.includeDownstreamBrowsers = true
var downstream = bbm.getCompatibleVersions(options)
options.includeDownstreamBrowsers = false
options.includeKaiOS = false
baselineVersions = bbm.getCompatibleVersions(options).concat(
  downstream.filter(function (version) {
    return version.browser === 'kai_os'
  })
)
```

`baseline 2020 including kaios` is now exactly `baseline 2020` plus
`kaios 3.0-3.1` — no `samsung`, `opera`, `and_uc`, `and_qq` or `android`. The
extra call only happens for this one query shape; every other query still makes
exactly one.

The four `getCompatibleVersions` call sites differed only in a single key, so
they are now one options object built up in the branches. That is what makes
the KaiOS case a small addition rather than a fifth copy.

`grammar.w3c-ebnf` keeps `with downstream` and `including kaios` independently
optional, which is now true, and moves both outside the year/availability
alternation — they had been attached only to the `newly`/`widely` branch, while
the regexp and the existing `baseline 2020 with downstream including kaios`
test already allowed them on both.

## A related limitation, not fixed here

`including kaios` only produces output for older baselines.
`baseline-browser-mapping` reports `kai_os 4.0` for anything from ~2021 on,
`caniuse-lite`'s `kaios` tops out at `3.0-3.1`, so `kaios >= 4.0` resolves to
nothing. That is true of `with downstream including kaios` on `main` today too
— `baseline widely available with downstream including kaios` contains no
`kaios` entry. It is a caniuse data gap rather than a query bug, so I left it
alone, but it is why the new tests use `baseline 2020`.

## Verification

- `Adds KaiOS and nothing else when downstream is not requested` asserts the
  set difference in both directions against `baseline 2020`: exactly
  `['kaios 3.0-3.1']` added, nothing dropped. On `main` + 2.11.12 it fails with
  a one-line diff (KaiOS missing).
- `Accepts "including kaios" without downstream in every baseline shape` covers
  all four shapes. On `main` + the locked 2.10.44 the suite does not fail, it
  **aborts** — uvu prints `Exiting early before testing is finished`.
- `unit` — 304 passed, against **both** `baseline-browser-mapping@2.10.44` and
  `@2.11.12`.
- `test:lint` — clean; `oxfmt --check` clean on both changed files.
- `test:coverage` — statements/lines/functions 100%. Branches 98.89%, the same
  as `main`, with the same two uncovered lines (375, 550).
- `test:size` — 24.96 kB against the 30 kB limit (24.89 kB on `main`).

README now documents `… including kaios` on its own rather than only as a
suffix of `… with downstream`. `CHANGELOG.md` left alone, matching #937/#938 —
happy to add an entry.

### DIFF
--- a/README.md
+++ b/README.md
@@ -231,7 +231,7 @@ You can specify the browser and Node.js versions by queries (case insensitive):
   - `baseline widely available on YYYY-MM-DD`: selects browser versions that supported the Widely available feature set on the specified date.
   - `baseline 2022`: selects browser versions that are compatible with all features that were Baseline Newly available at the end of the specified year.
   - `… with downstream`: includes browsers outside the core browser set that support the requested Baseline feature set based on their Chromium or Gecko version. See [`baseline-browser-mapping`](https://github.com/web-platform-dx/baseline-browser-mapping#downstream-browsers).
-  - `… with downstream including kaios`: same output as the previous query plus KaiOS.
+  - `… including kaios`: adds KaiOS, with or without `with downstream`.
 - Last versions:
   - `last 2 versions`: the last 2 versions for _each_ browser.
   - `last 2 Chrome versions`: the last 2 versions of Chrome browser.
--- a/grammar.w3c-ebnf
+++ b/grammar.w3c-ebnf
@@ -54,7 +54,7 @@ Years ::= 'last' Space+ Numeric Space+ 'year' 's'?
 
 Since ::= 'since' Space Digit+ ('-' Digit+ ('-' Digit+)?)?
 
-Baseline ::= 'baseline' Space (Digit+ | ('newly' | 'widely') Space 'available' (Space 'on' Space Digit+ ('-' Digit+ ('-' Digit+)?)?)? (' with downstream')? (' including kaios')?)
+Baseline ::= 'baseline' Space (Digit+ | ('newly' | 'widely') Space 'available' (Space 'on' Space Digit+ ('-' Digit+ ('-' Digit+)?)?)?) (Space 'with' Space 'downstream')? (Space 'including' Space 'kaios')?
 
 Region ::= ('alt-' [a-z][a-z] | [A-Z][A-Z])
 
--- a/index.js
+++ b/index.js
@@ -834,43 +834,44 @@ var QUERIES = {
     regexp:
       /^baseline\s+(?!\s)(?:(\d+)|(newly|widely)\s+(?!\s)available(?:\s+(?!\s)on\s+(?!\s)(\d{4}-\d{2}-\d{2}))?)?(\s+(?!\s)with\s+(?!\s)downstream)?(\s+(?!\s)including\s+(?!\s)kaios)?$/i,
     select: function (context, node) {
-      var baselineVersions
-      var includeDownstream = !!node.downstream
-      var includeKaiOS = !!node.kaios
       var availability = node.availability && node.availability.toLowerCase()
       if (availability === 'newly' && node.date) {
         throw new BrowserslistError(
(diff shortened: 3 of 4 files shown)


ai:
Why do you think we should not allow `baseline widely available including kaios`?

Maybe we should fix it?

Jaybhade:
You're right, it should work — I've rewritten the PR to support it.

I rejected it because `baseline-browser-mapping` refuses `includeKaiOS` without
`includeDownstreamBrowsers`, and I read that as "this combination is
meaningless". It isn't: KaiOS is downstream in *their* model, but "core
browsers plus KaiOS" is perfectly coherent from our side. So we now ask for the
whole downstream set, filter it down to KaiOS, and merge that onto the core
set. `baseline 2020 including kaios` is exactly `baseline 2020` plus
`kaios 3.0-3.1`, with no `samsung`/`opera`/`and_uc`/`and_qq`/`android`. The
second `getCompatibleVersions` call only happens for this one query shape.

While redoing it I found the symptom is version-dependent, which is worth
knowing either way:

| `baseline-browser-mapping` | `baseline widely available including kaios` |
| --- | --- |
| `2.10.44` (pinned in `pnpm-lock.yaml`, so what CI runs) | `process.exit(1)` |
| `2.11.12` (what `^2.10.44` resolves to today) | silently identical to `baseline widely available` |

2.11.x turned the exit into a real `Error`, but only in
`getAllVersions`/`getTimeline` — `getCompatibleVersions`, the one we call, just
drops KaiOS now. So on current npm the query is a silent no-op and on the
locked version it's fatal. Tests cover both; the suite is green against each.

One thing I did **not** fix, since it's a caniuse data gap rather than a query
bug: `including kaios` produces no output for modern baselines at all.
`baseline-browser-mapping` reports `kai_os 4.0` from ~2021 on, `caniuse-lite`'s
`kaios` stops at `3.0-3.1`, so `kaios >= 4.0` resolves to nothing — which is
already true of `with downstream including kaios` on `main`. That's why the new
tests use `baseline 2020`. Happy to look at it separately if you want.

ai:
Thanks. Released in 4.28.8.

PR #944, opened 2026-09-11
Support query continuations across lines and array entries

### LINKED ISSUE #757: Support "and" on multiple lines
As far as I can tell, there is no way to use an intersection of queries, i.e. "and", across more than one line in `.browserslistrc` or in separate indices of a `package.json` array.  The only exception is when using "not" since the "and" is implied.

When trying to construct a configuration using multiple intersections, this makes for a confusing and ugly read.  I would really like to see a more intuitive syntax supported across multiple lines, such as:

```
query[{, |or}|and][\n] query2...
```

Meaning, keep the same syntax for backwards compatibility except allow the same set of combiners at the end of a line which will apply to the next query.  Currently this throws an error as invalid.

fzlzjerry:
Fixes #757.

## Changes

Allow a trailing `and`, `or`, or comma to continue a query into the next config line or query-array entry. For example:

```text
last 2 versions and
> 0.5%
```

A linear grouping pass joins only explicit continuations before the existing clause parser runs. Unconnected entries remain implicit OR, composition stays left-to-right, and an incomplete final group is still parsed so missing right-hand queries produce an error. There are no dependency or browser-data changes.

The README documents both file and array forms. Eight regression groups cover AND/OR/comma, whitespace/case, implicit OR, AST composition, malformed input, real `.browserslistrc`/`package.json` loading, and configuration-section boundaries.

## Validation

- The new tests fail in five groups on the original implementation; all eight pass after the change.
- Full unit suite: 312 passed on Node 20/22/24/26; 311 passed on Node 14/16/18, matching the existing runtime-dependent test count plus these eight additions.
- Lint and coverage gates passed: 100% statements/functions/lines, 98.93% branches overall. Size: 25.21 kB against the 30 kB limit.
- A supplementary comparison of 5,184 deterministic IE query/composition forms matches the original parser's canonical single-line results. Four targeted grouping/error-path omissions are detected by the regression tests; restored source passes again.

Direct public-API smoke checks also pass on Node 8/10/12. With the locally installed dev dependencies, the full harness stops in uvu (`globalThis`, Node 8/10) or fs-extra (Node 12), identically on the original and changed source. Those are not counted as full-suite passes; the native CI setup remains a separate check.


ai:
Thanks. Released in 4.29.0.

PR #948, opened 2026-09-19
Fix case-insensitive negation queries

jakezwang:
Match the `not` prefix case-insensitively, as documented for queries. `browserslist('ie >= 9, NOT ie 11')` currently throws an unknown-query error instead of excluding IE 11. Original query text is preserved in the AST and error messages.

A combined regression covers mixed/uppercase negation, parsed output, comma/AND/array queries, and the error for a leading exclusion. The covered cases fail before the fix. Full tests pass on Node 16, 20, 22, 24 and 26; the Node 26 coverage, lint and size checks pass with 100% line coverage.

Review update: combined the case variants into one test. `pnpm test` passes (313 tests, 100% line coverage, lint and size checks).


test/parse.test.js
```diff
@@ -30,4 +30,30 @@ test('parses queries to AST', () => {
   ])
 })
 
+for (let operator of ['Not', 'nOt', 'NOT']) {
+  test(`parses ${operator} as a negation`, () => {
```
ai: Combine all variants into the single test

jakezwang: Combined the case variants and their parsing, query and error assertions into one test. `pnpm test` passes (313 tests, 100% line coverage, lint and size checks).

ai:
Thanks

PR #949, opened 2026-09-19
Make percentage and supports queries case-insensitive

wahidrizka:
Queries are case-insensitive (the README says so, and `cover` was made case-insensitive in #648), but four query types still match only lowercase keywords:

```js
browserslist('> 1% in US')      // works
browserslist('> 1% IN US')      // Unknown browser query `> 1% IN US`
browserslist('> 1% In ALT-AS')  // Unknown browser query
browserslist('> 10% In My Stats', { stats })                // Unknown browser query
browserslist('> 5% In browserslist-config-mycompany Stats') // Unknown browser query
browserslist('Supports es6-module')                         // Unknown browser query
browserslist('FULLY SUPPORTS css-grid')                     // Unknown browser query
```

Their regexps are the only ones with letters that miss the `i` flag, so this adds it. The `select` of `> 5% in US` already normalizes the region code, so `IN us` and `in ALT-AS` work once the query matches.

`supports` needed one more change: its `select` compares `node.supportType` with `'fully'` as written, so with the flag alone `FULLY supports x` would silently include partial support. It is now lowercased first, the same way `baseline` handles `availability` since #935.

The `country`, `custom`, `shareable-stats` and `feature` tests each get an "is case insensitive" test, like the other query types have. They fail without this change, and the `FULLY SUPPORTS` assertion also fails with the flag alone. The unit tests (316), coverage (100% lines), ESLint and size-limit (25.33 kB) pass.


ai:
Thanks!

PR #951, opened 2026-09-20
cleanup: Combine equivalent `not`/`and not` logic; implement `and` as opposite of `not`

charmander:
I’m not sure if there are any compatibility considerations with the first use of `Set` here; I noticed `Map` was already in use.


ai:
Thanks

